PR #20, opened 2015-04-23
Added tests and extra ajax functionality

assertchris:
@chillu @tractorcow

### DIFF
--- a/code/TagField.php
+++ b/code/TagField.php
@@ -8,34 +8,142 @@
  */
 class TagField extends DropdownField {
 	/**
+	 * @var array
+	 */
+	public static $allowed_actions = array(
+		'suggest',
+	);
+
+	/**
 	 * @var bool
 	 */
-	protected $readOnly;
+	protected $ajax = false;
+
+	/**
+	 * @var bool
+	 */
+	protected $readOnly = false;
+
+	/**
+	 * @var string
+	 */
+	protected $relationTitle = 'Title';
+
+	/**
+	 * @var int
+	 */
+	protected $ajaxItemLimit = 10;
 
 	/**
 	 * @var null|string
 	 */
-	protected $relationTitleField;
+	protected $recordClass;
 
 	/**
-	 * @param string      $name
+	 * @param string $name
 	 * @param null|string $title
-	 * @param array       $source
-	 * @param array       $value
-	 * @param bool        $readOnly
-	 * @param string      $relationTitleField
+	 * @param array $source
+	 * @param array $value
+	 * @param bool $readOnly
 	 */
-	public function __construct($name, $title = null, $source = array(), $value = array(), $readOnly = false, $relationTitleField = 'Title') {
-		$this->readOnly = $readOnly;
-		$this->relationTitleField = $relationTitleField;
+	public function __construct($name, $title = null, $source = array(), $value = array(), $readOnly = false) {
+		$this->setReadOnly($readOnly);
 
 		parent::__construct($name, $title, $source, $value);
 	}
 
 	/**
-	 * @param array $properties
+	 * @return bool
+	 */
+	public function getAjax() {
+		return $this->ajax;
+	}
+
+	/**
+	 * @param bool $ajax
 	 *
-	 * @return string
+	 * @return static
+	 */
+	public function setAjax($ajax) {
+		$this->ajax = $ajax;
+
+		return $this;
+	}
+
+	/**
+	 * @return bool
+	 */
+	public function getReadOnly() {
+		return $this->readOnly;
+	}
+
+	/**
+	 * @param bool $readOnly
+	 *
+	 * @return static
+	 */
+	public function setReadOnly($readOnly) {
+		$this->readOnly = $readOnly;
+
+		return $this;
+	}
+
+	/**
+	 * @return null|string
+	 */
+	public function getRelationTitle() {
+		return $this->relationTitle;
+	}
+
+	/**
+	 * @param string $relationTitle
+	 *
+	 * @return static
+	 */
+	public function setRelationTitle($relationTitle) {
+		$this->relationTitle = $relationTitle;
+
+		return $this;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function getAjaxItemLimit() {
+		return $this->ajaxItemLimit;
+	}
+
+	/**
+	 * @param int $ajaxItemLimit
+	 *
+	 * @return static
+	 */
+	public function setAjaxItemLimit($ajaxItemLimit) {
+		$this->ajaxItemLimit = $ajaxItemLimit;
+
+		return $this;
+	}
+
+	/**
+	 * @return null|string
+	 */
+	public function getRecordClass() {
+		return $this->recordClass;
+	}
+
+	/**
+	 * @param string $recordClass
+	 *
+	 * @return static
+	 */
+	public function setRecordClass($recordClass) {
+		$this->recordClass = $recordClass;
+
+		return $this;
+	}
+
+	/**
+	 * {@inheritdoc}
 	 */
 	public function Field($properties = array()) {
 		Requirements::css(TAG_FIELD_DIR . '/css/select2.min.css');
@@ -43,7 +151,6 @@ class TagField extends DropdownField {
 
 		Requirements::javascript(TAG_FIELD_DIR . '/js/TagField.js');
 		Requirements::javascript(TAG_FIELD_DIR . '/js/select2.js');
-
 		Requirements::javascript(THIRDPARTY_DIR . '/jquery/jquery.js');
 		Requirements::javascript(THIRDPARTY_DIR . '/jquery-entwine/dist/jquery.entwine-dist.js');
 
@@ -51,11 +158,41 @@ class TagField extends DropdownField {
 
 		$this->setAttribute('multiple', 'multiple');
 
+		if($this->ajax) {
+			$this->setAttribute('data-suggest-url', $this->getSuggestURL());
+		}
+
+		$properties = array_merge($properties, array(
+			'Options' => $this->getOptions()
+		));
+
+		return $this
+			->customise($properties)
+			->renderWith(array("templates/TagField"));
+	}
+
+	/**
+	 * @return string
+	 */
+	protected function getSuggestURL() {
+		return Controller::join_links($this->Link(), 'suggest');
+	}
+
+	/**
+	 * @return ArrayList
+	 */
+	protected function getOptions() {
 		$options = ArrayList::create();
+
+		$source = $this->getSource();
+
+		if($source instanceof Iterator) {
+			$source = iterator_to_array($source);
+		}
 
 		$values = $this->Value();
 
-		foreach(iterator_to_array($this->source) as $key => $value) {
+		foreach($source as $key => $value) {
 			$options->push(
 				ArrayData::create(array(
 					"Title" => $value,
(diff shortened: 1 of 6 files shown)


code/TagField.php
```diff
@@ -51,11 +158,41 @@ public function Field($properties = array()) {
...
+		return $this
+			->customise($properties)
+			->renderWith(array("templates/TagField"));
+	}
+
+	/**
+	 * @return string
+	 */
+	protected function getSuggestURL() {
+		return parse_url($this->Link(), PHP_URL_PATH) . '/suggest';
```
chillu: Please use Controller::join_links() - the controller might add GET params

code/TagField.php
```diff
@@ -173,4 +273,90 @@ public function saveInto(DataObjectInterface $record) {
...
+	 */
+	public function suggest(SS_HTTPRequest $request) {
+		if($this->recordClass === null) {
+			return '[]';
+		}
+
+		/**
+		 * @var DataObject $object
+		 */
+		$object = singleton($this->recordClass);
```
chillu: Why do we need this inline declaration? `singleton()` should already provide that level of type hinting. I also don't like block level comments inside methods, makes it harder to comment out stuff.

tests/travis/_config.php
```diff
@@ -0,0 +1,19 @@
+<?php
```
chillu: Is there a reason you didn't use https://github.com/silverstripe-labs/silverstripe-travis-support? Its standard across our other maintained modules, and allows for things like switching the core dependency easily.

tests/unit/TagFieldTest.php
```diff
@@ -0,0 +1,395 @@
...
+	 */
+	protected $extraDataObjects = array(
+		'TagFieldTest_BlogTag',
+		'TagFieldTest_BlogPost',
+	);
+
+	/**
+	 * @test
+	 */
+	public function itSavesLinksToExistingTagsOnExistingRecords() {
```
chillu: Wouldn't it be easier to just stick to the PHPUnit convention and prefix the methods with `test`? Then you don't need annotations for each of them. The sentence structure can stay the same.

tests/unit/TagFieldTest.php
```diff
@@ -0,0 +1,395 @@
...
+
+	/**
+	 * @var array
+	 */
+	private static $many_many = array(
+		'Tags' => 'TagFieldTest_BlogTag',
+	);
+}
+
+class TagFieldTest_Controller extends Controller {
```
chillu: Does this need a `TestOnly` as well?

code/TagField.php
assertchris: I added it to help with static analysis. `singleton` doesn't type-hint a class with `hasMethod` or `hasField` so a static analyser can't check expected types on that. Do you foresee a need to leave sections of this commented out? Perhaps that points to unnecessary code we could trim away?

code/TagField.php
```diff
@@ -173,4 +273,90 @@ public function saveInto(DataObjectInterface $record) {
...
+	 *
+	 * @return array
+	 */
+	protected function getObjectTags($class, DataObject $instance, $term) {
+		$relation = $instance->{$this->name}();
+
+		$query = DataList::create($relation->dataClass())
+			->where('"' . $this->relationTitle . '" LIKE \'%' . $term . '%\'')
+			->sort($this->relationTitle)
+			->limit($this->limit);
```
chillu: If we have defined getters, they should be used here - otherwise it makes it harder to overwrite functionality in a subclass

tests/travis/_config.php
assertchris: I copied this exact travis folder from `1.0`. Is there a reason you didn't use it? :P

code/TagField.php
```diff
@@ -173,4 +273,90 @@ public function saveInto(DataObjectInterface $record) {
...
+	 * @param DataObject $instance
+	 * @param string $term
+	 *
+	 * @return array
+	 */
+	protected function getObjectTags($class, DataObject $instance, $term) {
+		$relation = $instance->{$this->name}();
+
+		$query = DataList::create($relation->dataClass())
+			->where('"' . $this->relationTitle . '" LIKE \'%' . $term . '%\'')
```
chillu: Don't rely on method consumers remembering to escape SQL data (or put a big red warning into the PHPDoc about it).

code/TagField.php
```diff
@@ -173,4 +273,90 @@ public function saveInto(DataObjectInterface $record) {
...
+
+	/**
+	 * @param string $class
+	 * @param DataObject $instance
+	 * @param string $term
+	 *
+	 * @return array
+	 */
+	protected function getStringTags($class, DataObject $instance, $term) {
+		$column = '"' . $this->recordClass . '"."' . $this->name . '"';
```
chillu: Same here, needs escaping. Ideally using higher level methods though (haven't checked what's available on `DataQuery`)

tests/unit/TagFieldTest.php
assertchris: Might be nice to discuss as part of a larger unit test style guide. I don't think it matters either way...

tests/unit/TagFieldTest.php
assertchris: Copied the style of `1.0`. Do you want me to add that?

code/TagField.php
```diff
@@ -8,34 +8,142 @@
...
+	public function setRelationTitle($relationTitle) {
+		$this->relationTitle = $relationTitle;
+
+		return $this;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function getLimit() {
```
chillu: Unclear if this is the limit for auto-suggest, the allowed number of tags to be stored on a relation, or the number of tags loaded from an existing relation. Needs PHPDoc.

code/TagField.php
assertchris: Huh? https://github.com/silverstripe-labs/silverstripe-tagfield/blob/588530f09cda81a205166c1f7a39ccb934f6155e/code/TagField.php#L294

These methods are `protected` so it's really only subclasses that need to escape SQL data if they override `suggest`.

code/TagField.php
tractorcow: I'd rather refactor it into a method with the necessary type hinting in the return type. I also dislike the inline typehint.

tests/travis/_config.php
tractorcow: Check https://github.com/silverstripe/silverstripe-blog/blob/master/.travis.yml for an example on how we normally do this.

code/TagField.php
chillu: I thought `singleton()` was limited to `DataObject`, but that's not the case of course. OK, then then inline declaration makes sense. Stylistically, I think block level comments should only be used for class and method level docs. In order to debug, I sometimes use my IDE to select a bunch of code and comment out via keyboard shortcuts and inline comments (`//`). Its not blocking my workflow, just irks me to see it - you'll be hard pressed finding many block level comments inside methods in core, so its one of those "implicit conventions".

tests/unit/TagFieldTest.php
tractorcow: I never even knew about the markup. In any case, it's been convention up until now to prefix with test*

tests/travis/_config.php
chillu: lol, OK - that's reaaaaally old then. Might have actually been the first SilverStripe module ever to get travis support? Do you just want to copy the `.travis.yml` from another module and run with it? Much simpler: https://github.com/silverstripe/silverstripe-blog/blob/1.0/.travis.yml

tests/unit/TagFieldTest.php
chillu: Yes, please. I think `tests/*` classes get excluded from standard manifest builds, but it's still a nice safeguard that's used in core: https://github.com/silverstripe/silverstripe-framework/blob/3.1/tests/control/ControllerTest.php#L430

code/TagField.php
```diff
@@ -173,4 +273,90 @@ public function saveInto(DataObjectInterface $record) {
...
+	 * @return array
+	 */
+	protected function getStringTags($class, DataObject $instance, $term) {
+		$column = '"' . $this->recordClass . '"."' . $this->name . '"';
+		$value = '\'%' . $term . '%\'';
+
+		$query = new DataQuery($this->recordClass);
+
+		$query
+			->where('LOWER(' . $column . ') LIKE LOWER(' . $value . ')')
```
tractorcow: DataList already has extensions built in for case-sensitive filters.

```php
DataObject::get($this->RecordClass)
    ->filter("{$this->name}:PartialMatch:nocase", $value);
```

code/TagField.php
```diff
@@ -173,4 +273,90 @@ public function saveInto(DataObjectInterface $record) {
...
+	}
+
+	/**
+	 * @param string $class
+	 * @param DataObject $instance
+	 * @param string $term
+	 *
+	 * @return array
+	 */
+	protected function getStringTags($class, DataObject $instance, $term) {
```
tractorcow: Where is `$class` used?

code/TagField.php
chillu: > As a rule of thumb, you should escape your data as close to querying as possible. This means if you've got a chain of functions passing data through, escaping should happen at the end of the chain.
- http://docs.silverstripe.org/en/developer_guides/security/secure_coding

code/TagField.php
```diff
@@ -173,4 +273,90 @@ public function saveInto(DataObjectInterface $record) {
...
+		}
+
+		return $items;
+	}
+
+	/**
+	 * @param string $class
+	 * @param DataObject $instance
+	 * @param string $term
+	 *
```
tractorcow: PHPDoc description would be nice :D

code/TagField.php
```diff
@@ -173,4 +273,90 @@ public function saveInto(DataObjectInterface $record) {
...
+	 * @return array
+	 */
+	protected function getObjectTags($class, DataObject $instance, $term) {
```
tractorcow: See my comment on getStringTags below.

code/TagField.php
```diff
@@ -173,4 +273,90 @@ public function saveInto(DataObjectInterface $record) {
...
+
+		$query = DataList::create($relation->dataClass())
+			->where('"' . $this->relationTitle . '" LIKE \'%' . $term . '%\'')
+			->sort($this->relationTitle)
+			->limit($this->limit);
+
+		$items = array();
+
+		foreach($query as $record) {
+			$items[] = $record->{$this->relationTitle};
```
tractorcow: If just selecting one field, use `->column($this->relationTitle)` to only limit the SQL query to that one field.

code/TagField.php
tractorcow: And look no SQL injection risk there. :D

code/TagField.php
```diff
@@ -173,4 +273,90 @@ public function saveInto(DataObjectInterface $record) {
...
+
+		$tags = array();
+
+		if($object->hasMethod($this->name)) {
+			$tags = $this->getObjectTags($this->recordClass, $object, $term);
+		} elseif($object->hasField($this->name)) {
+			$tags = $this->getStringTags($this->recordClass, $object, $term);
+		}
+
+		return Convert::raw2json($tags);
```
tractorcow: Is it worth setting the content-type here? You can return a SS_HTTPResponse object instead with the necessary headers instead of a string return type.

code/TagField.php
assertchris: Does `:PartialMatch:nocase` apply in the case of `DataQuery`?

code/TagField.php
```diff
@@ -173,4 +274,123 @@ public function saveInto(DataObjectInterface $record) {
...
+	 *
+	 * @param SS_HTTPRequest $request
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function suggest(SS_HTTPRequest $request) {
+		$recordClass = $this->getRecordClass();
+
+		if($recordClass === null) {
+			return '[]';
```
tractorcow: You don't set the content type here.

code/TagField.php
```diff
@@ -173,4 +274,123 @@ public function saveInto(DataObjectInterface $record) {
...
+	protected function getObjectTags(DataObject $instance, $term) {
+		$name = $this->getName();
+		$relationTitle = $this->getRelationTitle();
+
+		$relation = $instance->{$name}();
+
+		$term = Convert::raw2sql($term);
+
+		$query = DataList::create($relation->dataClass())
+			->where('"' . $relationTitle . '" LIKE \'%' . $term . '%\'')
```
tractorcow: `->filter($relationTitle . ":PartialMatch", $term)` is the correct sql safe way to do this.

code/TagField.php
```diff
@@ -173,4 +274,123 @@ public function saveInto(DataObjectInterface $record) {
...
+			->sort($relationTitle)
+			->limit($this->getAjaxItemLimit());
```
tractorcow: Use `->column($relationTitle)` to simplify the SQL, meaning that DataList won't have to construct a separate DataObject instance for each row and MySQL won't have to return data for discarded columns.

code/TagField.php
```diff
@@ -173,4 +274,123 @@ public function saveInto(DataObjectInterface $record) {
...
+
+		$term = Convert::raw2sql($term);
+
+		$column = '"' . $recordClass . '"."' . $name . '"';
+		$value = '\'%' . $term . '%\'';
+
+		$query = new DataQuery($recordClass);
+
+		$query
+			->where('LOWER(' . $column . ') LIKE LOWER(' . $value . ')')
```
tractorcow: Same as above; Better to use filter api and ->column or ->map to only select the necessary fields.

assertchris:
Thanks @chillu and @tractorcow. I've implemented all the feedback.

chillu:
Thanks Chris!

PR #23, opened 2015-05-15
Stabilized API

assertchris:
This must be merged at the same time as https://github.com/silverstripe/silverstripe-blog/pull/244


tractorcow:
Should @chillu look at this one?

assertchris:
I would appreciate it, but your feedback is also good. :)

code/StringTagField.php
```diff
@@ -0,0 +1,283 @@
...
+	public function setLazyLoadItemLimit($lazyLoadItemLimit) {
+		$this->lazyLoadItemLimit = $lazyLoadItemLimit;
+
+		return $this;
+	}
+
+	/**
+	 * @return null|DataObject
+	 */
+	public function getDataObject() {
```
tractorcow: Should we call it getRecord / setRecord to be consistent with the Form API?

assertchris: You've got a point - maybe I should rename the property as well...

code/StringTagField.php
```diff
@@ -0,0 +1,283 @@
...
+		return $options;
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function setValue($value, $source = null) {
+		if($source instanceof DataObject) {
+			$name = $this->getName();
+			$value = $source->$name;
```
tractorcow: Shouldn't this split on comma as well? $value would be a simple text value at this point right?

code/StringTagField.php
```diff
@@ -0,0 +1,283 @@
...
+		return parent::setValue($value);
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function getAttributes() {
+		return array_merge(
+			parent::getAttributes(),
+			array('name' => $this->getName() . '[]')
```
tractorcow: That's the best way to do it :)

code/StringTagField.php
```diff
@@ -0,0 +1,283 @@
...
+		);
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function saveInto(DataObjectInterface $record) {
+		parent::saveInto($record);
+
+		if(!$record instanceof DataObject) {
```
tractorcow: Actually impossible since we are using the strongly typed parameter `DataObjectInterface`.

code/StringTagField.php
assertchris: Right!

code/TagField.php
```diff
@@ -55,89 +53,97 @@ public function __construct($name, $title = null, $source = array(), $value = ar
...
 		return $this;
 	}
 
 	/**
-	 * @return null|string
+	 * @return null|DataObject
 	 */
-	public function getRecordClass() {
-		return $this->recordClass;
+	public function getDataObject() {
```
tractorcow: Same comment as StringTagField. Should this be getRecord?

code/StringTagField.php
assertchris: /me shakes head

tractorcow: I actually had that backwards, sorry. :)

In any case, we should allow any object that implements that interface.

assertchris: I shook my head because of my redundant code. Not your comment. Your comment was spot-on.

code/TagField.php
```diff
@@ -17,37 +15,32 @@ class TagField extends DropdownField {
...
-	 * @param array $value
-	 * @param bool $readOnly
+	 * @param string $title
+	 * @param DataList $source
+	 * @param null|DataList $value
+	 * @param bool $canCreate
 	 */
-	public function __construct($name, $title = null, $source = array(), $value = array(), $readOnly = false) {
-		$this->setReadOnly($readOnly);
+	public function __construct($name, $title = '', $source, $value = null, $canCreate = true) {
```
tractorcow: You can't have non-optional arguments before optional ones.

code/TagField.php
```diff
@@ -17,37 +15,32 @@ class TagField extends DropdownField {
 	/**
 	 * @var bool
 	 */
-	protected $ajax = false;
```
tractorcow: None of these options (getters or setters) have phpdoc descriptions.

code/StringTagField.php
```diff
@@ -0,0 +1,283 @@
...
+	/**
+	 * {@inheritdoc}
+	 */
+	public function Field($properties = array()) {
+		Requirements::css(TAG_FIELD_DIR . '/css/select2.min.css');
+		Requirements::css(TAG_FIELD_DIR . '/css/TagField.css');
+
+		Requirements::javascript(TAG_FIELD_DIR . '/js/TagField.js');
+		Requirements::javascript(TAG_FIELD_DIR . '/js/select2.js');
+		Requirements::javascript(THIRDPARTY_DIR . '/jquery/jquery.js');
```
tractorcow: I'd adjust the order so that dependant scripts appear after their dependencies.

code/StringTagField.php
tractorcow: Hey, if I'm ever right about anything, it's about me being wrong. :D

code/TagField.php
assertchris: Not sure descriptions are needed for simple getters/setters... :/

chillu:
I'll have a look tonite

tractorcow:
@chillu I've reviewed and I'm happy to merge, but if you want final say I'll leave it to you.

code/StringTagField.php
```diff
@@ -0,0 +1,283 @@
+<?php
+
+/**
+ * @package forms
+ * @subpackage fields
+ */
+class StringTagField extends DropdownField {
```
chillu: Can you add some PHPDoc on the class purpose please?

code/StringTagField.php
```diff
@@ -0,0 +1,283 @@
...
+	public function Field($properties = array()) {
+		Requirements::css(TAG_FIELD_DIR . '/css/select2.min.css');
+		Requirements::css(TAG_FIELD_DIR . '/css/TagField.css');
+
+		Requirements::javascript(THIRDPARTY_DIR . '/jquery/jquery.js');
+		Requirements::javascript(THIRDPARTY_DIR . '/jquery-entwine/dist/jquery.entwine-dist.js');
+		Requirements::javascript(TAG_FIELD_DIR . '/js/select2.js');
+		Requirements::javascript(TAG_FIELD_DIR . '/js/TagField.js');
+
+		$this->addExtraClass('silverstripe-tag-field');
```
chillu: Around the CMS we have a `ss` prefix/suffix convention, rather than the full `silverstripe` name - maybe we should continue that here as well?

code/StringTagField.php
```diff
@@ -0,0 +1,283 @@
...
+
+		$source = $this->getSource();
+
+		if($source instanceof Iterator) {
+			$source = iterator_to_array($source);
+		}
+
+		$values = $this->Value();
+
+		foreach($source as $key => $value) {
```
chillu: The README example defines the `$source` as `array('one', 'two')`. You're looping over key/val here, which means the keys will be numeric. Yet in `suggest()` you're setting tags with `'id' => $tag`, so an associative array. Is the example wrong, or am I missing something here?

code/TagField.php
```diff
@@ -17,37 +15,32 @@ class TagField extends DropdownField {
...
-	 * @param array $value
-	 * @param bool $readOnly
+	 * @param string $title
+	 * @param null|DataList $source
+	 * @param null|DataList $value
+	 * @param bool $canCreate
 	 */
-	public function __construct($name, $title = null, $source = array(), $value = array(), $readOnly = false) {
-		$this->setReadOnly($readOnly);
+	public function __construct($name, $title = '', $source = null, $value = null, $canCreate = true) {
```
chillu: Do we really need `$canCreate` as a constructor arg? The FormField API is confusing enough without varying the constructor args :)

code/TagField.php
```diff
@@ -184,20 +159,37 @@ protected function getSuggestURL() {
 	protected function getOptions() {
 		$options = ArrayList::create();
 
+		/**
```
chillu: Still not happy with the block level comments inside a method body (I think Damian agreed last time I've brought this up). Does your IDE ignore the hint if its defined through a `//` comment?

code/TagField.php
```diff
@@ -241,108 +231,110 @@ public function saveInto(DataObjectInterface $record) {
...
+		}
+
+		if(empty($record) || empty($source) || empty($titleField)) {
 			return;
 		}
 
+		/**
+		 * @var DataObject $record
+		 */
 		if($record->hasMethod($name)) {
```
chillu: Should this throw an exception if no method can be found? Otherwise the field save won't have any effect, leading to frustrating debugging sessions?

chillu:
Apart from a few naggy bits, great job - good test coverage as well :)

PR #26, opened 2015-06-11
Added foreign key so that it doesn't get categories and tags from elsewhere

pitchandtone:


code/TagField.php
```diff
@@ -310,7 +310,10 @@ protected function getTags($term) {
 		$term = Convert::raw2sql($term);
 
 		$query = $dataClass::get()
-			->filter($titleField . ':PartialMatch:nocase', $term)
+			->filter(array(
```
tractorcow: My understanding is that HasManyList should already include this filter.

Also, you run the risk where $source is actually a DataList, not a *ManyList, and thus won't have a method `getForeignKey`.

Did you encounter this issue where tags / categories from other blogs were actually being included?

code/TagField.php
```diff
@@ -310,7 +310,10 @@ protected function getTags($term) {
 		$term = Convert::raw2sql($term);
```
tractorcow: Lol, I just noticed, this shouldn't be there. :)

tractorcow:
Some comments; @assertchris  (aka tagfield master) any feelings?

code/TagField.php
assertchris: Good to know.

code/TagField.php
assertchris: `TagField` is meant to be able to work with any `DataList`. The source tags depend on the source `DataList` that you provide the constructor/factory...

pitchandtone: Since the code recreates the query, the filter is no longer there. The code just grabs the dataClass but not any of the filter params. So yes, it picks up Tags from other blogs across the site.

So either the code needs to add the terms filter to the existing dataSource or grab the filter from the dataSource provided.

While I appreciate that it could take any kind of DataList, it seems like that's an edge case and the original desired functionality of grabbing the tags from a specific blog is not working.

Can someone guide me about how you would like it impleented:
a) Use the original datasource provided and adda filter (unsure if this will work...)
b) Check what kid of datalist is provided and if it's a Has Many then do the code provided.

dhensby:
@pitchandtone can you rebase, please?

dhensby:
Ping @assertchris

code/TagField.php
tractorcow: > Since the code recreates the query

This was fixed in a recent update; It no longer does this. :)

tractorcow:
I'm not sure the problem this code solves exists anymore. It should have been fixed with https://github.com/silverstripe-labs/silverstripe-tagfield/commit/75a7e956ba3d66b35ac38bab76d8a691bd79e9cc

Close?

tractorcow:
Fixed with https://github.com/silverstripe-labs/silverstripe-tagfield/pull/57

Sorry @pitchandtone for re-fixing the issue you had already made a good attempt at. Still appreciate your work. ;D

PR #27, opened 2015-07-07
BUGFIX: Allow use on frontend

lingo:
'Chosen' plugin assumed, but on front-end this is not
a valid assumption.


assertchris:
Good idea! Just need to check it locally, but looks good so far...

assertchris:
Thanks!

PR #32, opened 2015-07-30
Extension Usage

danrye:
Curious what you think of this usage, would you consider adding this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,25 @@
 
 ## Usage
 
+### Relational Tags w/ Extension
+
+```php
+class Tag extends DataObject {
+    static $db = array(
+        'Title' => 'Varchar(200)',
+    );
+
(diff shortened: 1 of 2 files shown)


tractorcow:
I don't think it would work because `Tag` isn't a class that's included in this module, I'm afraid. :P

danrye:
Right, they had to create that before, but it minimizes what a dev has to bring to the table.  Just a thought.  You could include tag, but then they'd have to extend tag to implement the belongs_many and tag is so simple it didn't seem worth it. 



—

dhensby:
I think including this in the docs is helpful, but including the Extension doesn't make sense if the dependencies don't exist.

dhensby:
:+1:

dhensby:
No movement on this in 1.5 years - closing.

PR #33, opened 2015-08-10
examples using create and chaining

barry-bookinglive:
New silverstripe funcitonality of create and chaining to be highlighted in the example


dhensby:
Thanks

PR #34, opened 2015-08-17
FIX: Override DropdownField::validate for SS3.2 compatiblity

markguinn:
Prior to this fix a clean install of 3.2 + blog yields the following error when saving a post:

```
ERROR [Warning]: array_key_exists(): The first argument should be either a string or an integer
IN POST /admin/pages/edit/EditForm
Line 323 in /Users/guinns/projects/blog-test/framework/forms/DropdownField.php
```

### DIFF
--- a/code/TagField.php
+++ b/code/TagField.php
@@ -327,4 +327,17 @@ class TagField extends DropdownField {
 
 		return $items;
 	}
+
+
+	/**
+	 * DropdownField assumes value will be a scalar so we must
+	 * override validate. This only applies to Silverstripe 3.2+
+	 *
+	 * @param Validator $validator
+	 * @return bool
+	 */
+	public function validate($validator) {
+		return true;
+	}
+
 }


dhensby:
Thanks for digging into this and finding a solution.

Though I feel this validation function doesn't actually provide the value it should (ie: making sure only legitimate values have been selected). Perhaps the problem here is that the `value` of the tag field is an array?! Maybe the validate function should be closer to `CheckboxSetField`?

markguinn:
I wondered about that. I don't feel like I have enough understanding of how the field works, especially related to creating new values. I can give it a deeper look tomorrow. It looks to me like the 3.1 version didn't have any validation though, so this seemed like we're at least not losing anything between 3.1 and 3.2.

tractorcow:
I think for the sake of this module, this is probably ok. The current validate is broken and needs to be overridden, and should accept un-bound options as inputs anyway (since these are created if they don't exist).

Issue #30, opened 2015-07-28
Warning at \...SSViewer.php line 788: None of the following templates could be found (no theme in use): templates/StringTagField.ss 


barry-bookinglive:
I get the following warning when trying to use the StringTagField...

Warning at \framework\view\SSViewer.php line 788: None of the following templates could be found (no theme in use): templates/StringTagField.ss

3Dgoo:
Yes, there is no StringTagField template. I've tried changing this to TagField.ss. This seems to fix that problem, although other problems pop up with StringTagField later on. 

Should StringTagField be calling _templates/TagField.ss_ instead of _templates/StringTagField.ss_ ? Or is there meant to be a _StringTagField.ss_ file?

tractorcow:
@assertchris know if this is a quick fix?

PR #36, opened 2015-08-21
Fixing missing template error

3Dgoo:
The StringTagField tries to use a StringTagField.ss template, which doesn't existing. I have changed this to use the TagField.ss template instead.

This fixes issue #30


assertchris:
Thanks

PR #39, opened 2015-09-03
Upgrading select2 code to the latest version.

3Dgoo:
This fixes a rendering bug in safari where the hidden text box has a border displayed around it, making it not so hidden.

tag-field-safari-bug


assertchris:
Thanks for trying to fix this, but it would be much better to make the fix in `TagField.css` instead of `select2.css`. Also +1,148 −470..?

3Dgoo:
It's all good Chris. Sorry for bombarding the module with pull requests lately. 

Yeah I'm not 100% sure on all the things they updated in select2, but looking at the [repositories log on github](https://github.com/select2/select2) I trust they only let through fixes they've thoroughly tested. Hopefully it's all good.

assertchris:
No need to be sorry - pull requests are great! 

I'm sure it's 100%, but I need to test it or @tractorcow will roast me... :D

tractorcow:
I only roast you if you don't test things and they break. Just make sure to not test things that aren't broken and you'll be fine!

assertchris:
Thanks!

PR #46, opened 2015-10-06
BUGFIX: Fixed notice from array_filter when the value is null

UndefinedOffset:
In cases where the value of the tag field is null a notice is thrown (see bellow) this pull hands off to DropdownField's setValue() when the value is not an array.

```
array_filter() expects parameter 1 to be array, null given tagfield/code/TagField.php:210
```


assertchris:
Thanks

PR #56, opened 2015-10-24
Check if a Tag already exists on save

Zauberfisch:


tractorcow:
Good solution, but it should query the source to see if it exists rather than just looking at the dataclass.

See https://github.com/silverstripe-labs/silverstripe-tagfield/pull/57

dhensby:
@Zauberfisch this needs rebasing

@tractorcow was this fixed by another PR?

PR #57, opened 2015-11-03
BUG Fix items not included from source appearing in suggestions

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-blog/issues/295

This is because multiple blogs might have overlapping tags.

### DIFF
--- a/code/TagField.php
+++ b/code/TagField.php
@@ -332,13 +332,11 @@ class TagField extends DropdownField {
 		 */
 		$source = $this->getSource();
 
-		$dataClass = $source->dataClass();
-
 		$titleField = $this->getTitleField();
 
 		$term = Convert::raw2sql($term);
 
-		$query = $dataClass::get()
+		$query = $source
 			->filter($titleField . ':PartialMatch:nocase', $term)
 			->sort($titleField)
 			->limit($this->getLazyLoadItemLimit());
--- a/tests/TagFieldTest.php
+++ b/tests/TagFieldTest.php
@@ -148,6 +148,44 @@ class TagFieldTest extends SapphireTest {
 	}
 
 	/**
+	 * Tests that TagField supports pre-filtered data sources
+	 */
+	public function testRestrictedSuggestions() {
+		$source = TagFieldTestBlogTag::get()->exclude('Title', 'Tag2');
+		$field = new TagField('Tags', '', $source);
+
+		/**
+		 * Partial tag title match.
+		 */
+		$request = $this->getNewRequest(array('term' => 'Tag'));
+
+		$this->assertEquals(
+			'{"items":[{"id":1,"text":"Tag1"}]}',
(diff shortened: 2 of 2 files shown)


mschilder:
Hi! This fixed solved the issue when using the MOUSE to start the seach for tags (the search will result nothing). However if you start TYPING the tag (if you know the tag name), the tagfield will still display and accept tags from another BlogHolder. After save the table BlogCategories is altered: BlogID has changed from the initial BlogHolder to the new BlogHolder.

tractorcow:
I've done a test and it appears that select / typing brings up the correct list.

Can you please confirm you're testing against the current master (tagfield: b5cf2a8d27d0d13ced8d54cb0203665592cf0eb6, blog: 3ef14a6e1a1f843ee994ac7e3678fb507bb29a37)

Are the tags from the other blog coming up in the suggestions, or does it only appear if you type the entire tag?

Thanks buddy and appreciate your review. :D

mschilder:
Hi, have downloaded latest tagfield ZIP. Not sure about blog. Please see this screencast. Current BlogHolder only has "Geluk" as BlogCategory and not "Vlogs" (that BlogCategory belongs to another Blog(Holder)).
https://dl.dropboxusercontent.com/u/85464/screencast.mov

mschilder:
Thanks! Just gave this a try myself and see something different happening :) Created another video where Blog(Holder) /blog has ID=6 and Blog(Holder) /vlogs has ID=802. The new BlogPost I'm working on is a child of /vlogs. See what happens....
https://dl.dropboxusercontent.com/u/85464/screencast-2.mov

tractorcow:
Hi @mschilder , I think that's right. :D The category gets the BlogID of the holder, and there will be another table `BlogPost_Categories`. That will have the ID of the post rather than the holder.

It's a slightly de-normalised db schema, but it's necessary because sometimes categories might not be assigned to blogposts at any one point in time.

mschilder:
Thanks for update! However shouldn't this be more correct:

- If you type in a category that exists on another BlogHolder (and save that BlogPost)
- Then DO NOT change column BlogID in table Blog_Category for that category
- Instead, create a new row in table Blog_Category with the same name, but with new BlogID

tractorcow:
The problem is that there is no Blog_Category mapping, only BlogPost_Category.

I suppose that we COULD change the relationship between blog and category to many_many from has_many, but that would be a new major release.

mschilder:
We'll best to leave it like this than. I think this is an exception really :) Thanks for clearing this out for me though!!!

tractorcow:
It's worth discussing. That way we can see weaknesses in the current implementation.

Moving it to many_many might be beneficial in the long run.

dhensby:
IMO you could move to many_many in a BC way by adding the polyfill functions

PR #58, opened 2015-11-04
BUG Fix duplicate tag creation

tractorcow:
BUG Fix isMultiple not being initialised
BUG Fix tag rendering as a 100px wide container

### DIFF
--- a/code/TagField.php
+++ b/code/TagField.php
@@ -35,9 +35,9 @@ class TagField extends DropdownField {
 	protected $titleField = 'Title';
 
 	/**
-	 * @var string
+	 * @var bool
 	 */
-	protected $isMultiple;
+	protected $isMultiple = true;
 
 	/**
 	 * @param string $name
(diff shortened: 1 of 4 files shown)


css/TagField.css
```diff
@@ -1,6 +1,7 @@
 .select2-container {
     position: relative;
     max-width: 416px !important;
+    width: 100%;
```
tractorcow: I think this is necessary if the field is rendered inside a container that's hidden on load.

code/TagField.php
```diff
@@ -334,28 +355,24 @@ protected function getTags($term) {
 
 		$titleField = $this->getTitleField();
 
-		$term = Convert::raw2sql($term);
```
tractorcow: Because we shouldn't encode anything that's passed into the `filter` API.

code/TagField.php
```diff
@@ -334,28 +355,24 @@ protected function getTags($term) {
...
-
+		$titleField = $this->getTitleField();
 		foreach($query->map('ID', $titleField) as $id => $title) {
-			if(!in_array($title, $items)) {
-				$items[] = array(
-					'id' => $id,
-					'text' => $title
-				);
-			}
+			$items[$title] = array(
```
tractorcow: Noticed the old code was a bit weird so I changed the uniqueness check.

PR #67, opened 2015-11-23
Fixing setValue null value bug

3Dgoo:
On rare occasions when a form is submitted `StringTagField` throws the following error at the end of the `setValue` function: 

    ERROR [Warning]: array_filter() expects parameter 1 to be array, null given

I have added some code that checks if `$value` is `null`. If it is `null` it sets `$value` to an empty array to prevent the error from occurring.

### DIFF
--- a/code/StringTagField.php
+++ b/code/StringTagField.php
@@ -223,6 +223,10 @@ class StringTagField extends DropdownField
             $value = $source->column('ID');
         }
 
+        if (is_null($value)) {
+            $value = array();
+        }
+
         return parent::setValue(array_filter($value));
     }
 


tractorcow:
I don't think these test errors are your fault; The tests have hard-coded IDs which is going to break, obviously, on db schemas which don't truncate their ID auto-increment after each test. :)

3Dgoo:
Thanks for that info Damian. I was wondering why my pull requests always fail the build tests. Good to know it's not me.

3Dgoo:
Is there anything I can do to help this get merged in?

3Dgoo:
Thanks Damian.

PR #68, opened 2015-12-31
Bugfix: show source tags if not lazyloading

### LINKED ISSUE #25: No Tags available if not lazy loading
When not Lazy Loading, for pages with tags, the list displays all the avilable tags.  For new pages with no tags the list is empty.  Turning on lazy loading fixes this. 

````
    $field = new TagField(
      'RelationTags', 'Nursing, Corporate etc', Tag::get(), $this->RelationTags()
    );
//    $field->setShouldLazyLoad(true); // uncomment to workaround
    $field->setCanCreate(true);      // new tag DataObjects can be created
````

micschk:
Fixes #25: no source tags would be shown if not lazy loading (on new records/records without any pre-selected tags)

### DIFF
--- a/code/TagField.php
+++ b/code/TagField.php
@@ -208,15 +208,16 @@ class TagField extends DropdownField
 
         $values = $this->Value();
 
-        if (!$values) {
-            return $options;
+        // Mark selected tags while still returning a full list of possible options
+        $ids = array(); // empty fallback array for comparing
+        $values = $this->Value();
+        if($values){
+            // @TODO conversion from array to DataList to array...(?)
+            if(is_array($values)) {
+                $values = DataList::create($dataClass)->filter('ID', $values);
+            }
+            $ids = $values->column('ID');
         }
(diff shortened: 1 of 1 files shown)


tractorcow:
Good fix, but the tests are broken due to reliance on hard coded DB ids (which is a flaw in the test design, to be honest).

In any case, this change will need a unit test. :) How confident do you feel in writing one?

You can soft-code the test fixture ids with `$id = $this->idFromFixture('TagFieldTestBlogTag', 'Tag1)`; etc

micschk:
Writing tests is on my should-be-doing-when-i-find-the-time list, I'll add
this one :)

So, I'll do it eventually -- may take a while though (sorry).

dhensby:
assuming #70 passes and gets merged, this will either need a re-base or the CI tests restarting

dhensby:
tests pass now that #70 is merged

PR #70, opened 2016-01-05
FIX Tests with hard coded IDs

dhensby:
Fixes the tests that rely on tags having specific IDs


dhensby:
whoop tests pass @tractorcow

tractorcow:
Oh thanks @dhensby !

PR #71, opened 2016-01-27
Removed is_numeric check in SaveInto function

mediabeastnz:
I had an issue where saving integer only tags against a many_many relation wouldn't work but values like "A81" would.

There was a is_numeric check in the saveInto function which was prevent values like "123" being saved.

Is there an issue with having integers? If not please merge ;)

### DIFF
--- a/code/TagField.php
+++ b/code/TagField.php
@@ -231,7 +231,7 @@ class TagField extends DropdownField
         }
 
         return $options;
- 	}
+     }
 
     /**
      * {@inheritdoc}
@@ -310,7 +310,7 @@ class TagField extends DropdownField
 
         $relation->setByIDList(array_filter($ids));
 
- 	}
+     }
 
     /**
      * Get or create tag with the given value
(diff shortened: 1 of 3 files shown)


tractorcow:
What happens if you re-save a tagfield without editing it? Won't you now end up with a set of tags where the Title = the numeric ID?

mediabeastnz:
Ahh yes your right it does then save the tag with a title of the numeric ID. Can you think of a way to allow just numeric tags?

Also i'll create an issue but if you currently enter a tag that is just numeric it doesn't save a tag in the has_one table but it does create a record in the many_many table that has a TagID of the title you entered e.g. 123. Issue https://github.com/silverstripe-labs/silverstripe-tagfield/issues/72

tractorcow:
See my comment and possible solution at https://github.com/silverstripe-labs/silverstripe-tagfield/issues/72#issuecomment-179008130

tractorcow:
The updated fix looks good.

I think you need to fix your indentation settings; We are using psr-2, so you'll need to make sure your IDE respect the editorconfig (spaces not tabs for indentation).

mediabeastnz:
Should be good now, thanks.

tractorcow:
It looks like you need to update the tests; They are still expecting numeric IDs in the json response.

mediabeastnz:
Oh right. I'll update those tests now to allow for this change.

tractorcow:
Thanks. :D looking forward to it.

Honestly, I think the way you're making it work is the way it probably should have been from the start.

mediabeastnz:
Has anyone had a chance to review this?

tractorcow:
@mediabeastnz swap out the `$tag->ID` with `$tag->Title`. Remember we aren't using IDs anymore.

stevie-mayhew:
@mediabeastnz have you had time to review the feedback above? 😄

albanyacademy:
@mediabeastnz also looking forward to this fix :)

tractorcow:
@mediabeastnz here.

```php
$field = new TagField('Tags', '', new DataList('TagFieldTestBlogTag'), array($tag->ID, 'Tag3'));
```

Becomes

```php
$field = new TagField('Tags', '', new DataList('TagFieldTestBlogTag'), array($tag->Title, 'Tag3'));
```

tractorcow:
I'm glad I caught up on all my unreads... I'm getting better at responding to issues again. :)

PR #81, opened 2016-07-21
Apply canCreate permissions to Select2 for tags

### LINKED ISSUE #80: Improve the "can create" API
Hi guys,

Nice module!

I went to add a permission for "create new tags" today and grant it for admins but not for another role.

I see that the follow behaviours exist:

* Regardless of what `TagField::getCanCreate` returns, a new tag is always shown in the select when entered
* If the permission is to deny, the user can still select the new tag, and it is still added to the select area
* On saving the Page (or whatever the TagField is in) the new tag is discarded

It'd be great if we could not show new tags if you don't have permission to create them!

Cheers,
Robbie

robbieaverill:
* Resolves #80

### DIFF
--- a/code/StringTagField.php
+++ b/code/StringTagField.php
@@ -164,6 +164,8 @@ class StringTagField extends DropdownField
             ));
         }
 
+        $this->setAttribute('data-can-create', (int) $this->getCanCreate());
+
         return $this
             ->customise($properties)
             ->renderWith(array("templates/TagField"));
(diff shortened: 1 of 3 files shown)


robbieaverill:
*Bump* @tractorcow :)

assertchris:
I'll check this out...

assertchris:
Looks good, thanks! :)

tractorcow:
Sorry, I'm still catching up on my backlog of issues... this fix looks great. :D

PR #83, opened 2016-08-21
added missing cancreate methods and property to StringTagField.

sanderha:
``` $this->setAttribute('data-can-create', (int) $this->getCanCreate()); ``` would cause an error on StringTagField, because ``` getCanCreate() ``` was not defined.


dhensby:
Looks legit, thanks!

assertchris:
The amount of duplication between these classes is dumb. I am to blame. A 4.0-compatible version should make use of traits or composition...

PR #90, opened 2017-01-19
Fix for if the value is empty

oilee80:
Otherwise no options are returned and the dropdown is empty


oilee80:
Was wrongly implemented, closing

PR #100, opened 2017-10-18
FIX Update module for latest SS4, vendorise

robbieaverill:


robbieaverill:
Self-merging after @NightJar's review

PR #108, opened 2018-02-08
Tag field to look similar to list field

fonsekaean:
This contains bits of CSS and JS updates so that the tag fields looks similar to the ListBox fields and other fields in the CMS.

### DIFF
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -7,6 +7,7 @@ use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Forms\DropdownField;
+use SilverStripe\Forms\ReadonlyField;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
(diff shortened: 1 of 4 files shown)


andrewandante:
@fonsekaean You'll need to use the correct class for `ReadonlyTagField` - and a couple of screenshots to show the change would be great :)

src/TagField.php
```diff
@@ -446,7 +450,7 @@ public function validate($validator)
      */
     public function performReadonlyTransformation()
     {
-        $copy = $this->castedCopy(ReadonlyTagField::class);
```
andrewandante: This is the correct class, please rebase or undo this change

fonsekaean: Yup thats fixed ;)

fonsekaean:
Here is a screenshot of the cnage and how it looks like  
screen shot 2018-02-19 at 6 30 09 pm

andrewandante:
Tests need fixing, but screenshot looks good, thanks!

fonsekaean:
@andrewandante Yup those Travis issues are fixed

andrewandante:
Excellent, happy to ignore the codecov failures. Can you rebase/squash down to one commit please, and then I'll get someone with merge-powers to pull it in?

dhensby:
squash and merge feature is a life saver

PR #109, opened 2018-04-06
Set "title field" at a meaningful point in TagField instantiation

jakxnz:
* Added parameter to SilverStripe\TagFieldTagField::__construct()
* Modified SilverStripe\TagFieldTagField::getOptions() to use assigned title field
* Modified SilverStripe\TagFieldTagField::getValue() to use assigned title field

### DIFF
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -66,9 +66,10 @@ class TagField extends DropdownField
      * @param null|DataList $source
      * @param null|DataList $value
      */
-    public function __construct($name, $title = '', $source = [], $value = null)
+    public function __construct($name, $title = '', $source = [], $value = null, $titleField = 'Title')
     {
         $this->setSourceList($source);
+        $this->setTitleField($titleField);
         parent::__construct($name, $title, $source, $value);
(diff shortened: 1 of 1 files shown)


src/TagField.php
```diff
@@ -66,9 +66,10 @@ class TagField extends DropdownField
      * @param null|DataList $source
      * @param null|DataList $value
```
robbieaverill: Could you please add the new $titleField argument here?

src/TagField.php
```diff
@@ -283,10 +284,10 @@ public function setValue($value, $source = null)
             $name = $this->getName();
 
             if ($source->hasMethod($name)) {
-                $value = $source->$name()->column('Title');
+                $value = $source->$name()->column($this->titleField);
             }
         } elseif ($value instanceof SS_List) {
-            $value = $value->column('Title');
+            $value = $value->column($this->titleField);
```
robbieaverill: Could you please use `$this->getTitleField()` here, so we dogfood our own API?

PR #113, opened 2018-07-16
Update module to use react-select

### LINKED ISSUE #110: Upgrade TagField for SS4 to use react-select
We use [react-select](https://jedwatson.github.io/react-select/) in silverstripe/admin for fields like TreeDropdownField.

We should replace the chosen implementation with an example like this:

image

ScopeyNZ:
Resolves #110

This PR implements the change requested in the above issue. I added yarn and updated all the directory structure to match what new SS4 modules usually do. This is a backwards compatible enhancements so can me released in a minor - I have set the version in `package.json` to 2.2.0 - for what that's worth.

I noticed the implementation on the `disabled` attribute on the `react-select` dependency isn't great - might want to look at raising an issue for that.


ScopeyNZ:
Ok Travis pointed out that I lied, I've changed the format of the suggestion API on tagfield. I could change this to be backwards compatible?

client/src/components/TagField.js
```diff
@@ -0,0 +1,96 @@
...
+  getOptions(input) {
+    if (!this.props.lazyLoad) {
+      return Promise.resolve({ options: this.props.options });
+    }
+
+    if (!input) {
+      return Promise.resolve({ options: [] });
+    }
+
+    const fetchURL = url.parse(this.props.optionUrl, true);
```
robbieaverill: Could you please dereference these props - lazyLoad, options and optionUrl? There's also some places in the render method that could use this too

client/src/components/TagField.scss
```diff
@@ -0,0 +1,36 @@
...
+
+.is-focused:not(.is-open)>.Select-control,
+.is-open>.Select-control,
+.Select-menu-outer {
+  box-shadow: none;
+  border-color: $brand-primary;
+}
+
+.Select-menu-outer {
+  border-top-color: #e6e6e6;
```
robbieaverill: Is there a variable in admin somewhere we can use here?

package.json
```diff
@@ -20,17 +20,54 @@
...
+    ]
+  },
   "devDependencies": {
     "@silverstripe/eslint-config": "^0.0.4",
     "@silverstripe/webpack-config": "^0.8.0",
-    "expose-loader": "^0.7.4",
-    "sass-loader": "^6.0.7"
+    "@storybook/addon-actions": "^3.2.19",
+    "@storybook/addon-options": "^3.2.19",
+    "@storybook/react": "^3.2.19",
```
robbieaverill: Were you going to add a story book definition for it? That'd be cool, but not required

ScopeyNZ: I was playing around with that but I didn't get around to it. I can remove that.

ScopeyNZ:
I removed the BC concerns so this can target the next minor now 🙂

pxwee5:
`This is a backwards compatible enhancements so can me released in a minor`

If it's backwards compatible, can you make the silverstripe/framework dependency abit lower than 4.3 ?

Issue #115, opened 2018-09-06
TagField Performance with large number of tags


caffeineinc:
`TagField` is used in SilverStripe Blog to reference tags, and autocomplete the list. However with a large number of total tags (~12000) Performance in the CMS is hindered due to a parent class hydrating the entire tag list for the form. 

`TagField` is a subclass of `DropdownField`, and `SelectField` and all descendants of FormFields must call `parent::__construct()`. Unfortunately this means that `SelectField::getListMap()` is called from setSource in the constructor of SelectField which transform the `HasManyRelation` (set as the source) and hydrates it into an array (Effectively returning all 12000 tags, and storing them locally in that object). 

For a SelectField that makes sense as you need the source to determine if a value should be allowed. `TagField` handles both states, both with and without source items.  If lazy load is used, the source shouldn't need the reference hydrated as it submits a request for this from the frontend (and a check on submission). 

The problem is this happens during construction, and lazy load could be applied at a later time, thus there's no easy way of assuming it other than using `sourceList` over the parent `source` - It makes sense that these objects should only be hydrated when actually used to avoid performance problems.

Are there any other suggestions towards resolving that problem? 

The fix would be changing the constructor to send an empty source array:  instead of the source
```
public function __construct($name, $title = '', $source = [], $value = null, $titleField = 'Title')
    {
        $this->setSourceList($source);
        $this->setTitleField($titleField);

//        parent::__construct($name, $title, $source, $value);

        parent::__construct($name, $title, [], $value);
    }
```

caffeineinc:
`It works` 👍  haha 

Unfortunately i don't think a change would fit into semver, as the only way to change it would be omitting the source from the parent, and that would potentially break other behaviour that relies on the source. 

The component would have to know about lazy loading at instantiation, which means changing the interface or maybe splitting the class into a two components (lazyload tagfield and non-lazy tagfield). 

Happy to provide a PR but would like to establish direction before doing so.

ScopeyNZ:
I suppose you could limit the breakage by extending `getSource`:

```
public function getSource()
{
    if (!$this->source) {
        $this->setSource($this->getSourceList());
    }
    return $this->source;
}
```

But you're right - anyone relying on `->source` would be affected.

caffeineinc:
Yeah, further analysis also shows `TagField` mapping the entire `sourceList` for flagging and displaying the `checked` items  (it renders the entire tree :/) 

[...]

PR #116, opened 2018-09-14
Resolve TagField performance for lazyload

caffeineinc:
As a descendent of SelectField when source is passed to the parent, `getListMap()` hydrates any objects in the list. For a ManyMany or a HasMany relationship that can be extremely costly, as these could contain 1000's of elements. 

Instead we pass an empty array, and hydrate it on demand if requested. If the TagField uses lazy load, we don't need to hydrate the RelationList for all possible options as these can be loaded via XHR and better limited later. Instead, we only render the values and the options for those values into the component itself, and everything else can be done via separate requests.

@see #115

### DIFF
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -71,7 +71,7 @@ class TagField extends DropdownField
     {
         $this->setSourceList($source);
         $this->setTitleField($titleField);
-        parent::__construct($name, $title, $source, $value);
+        parent::__construct($name, $title, [], $value);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/TagField.php
```diff
@@ -298,6 +312,21 @@ public function setValue($value, $source = null)
...
+    /**
+     * Gets the source array if required
+     *
+     * Note: this is expensive for a SS_List
+     *
+     * @return array
+     */
+    public function getSource()
+    {
+        if (is_null($this->source)) {
```
ScopeyNZ: Above you overwrite the default `$source` to just an empty array but you're checking for null here. This means consumers of `getSource()` will now just get empty arrays?

tractorcow: My fix addresses this in setSource (it'll set to null if we have a DataList).

robbieaverill:
Cool, the commits from this PR are now in #117

PR #117, opened 2018-09-17
Fix lazy tags

tractorcow:
Added extra semver fixes to https://github.com/silverstripe/silverstripe-tagfield/pull/116. Now setSource() is supported and can be used to populate getSourceList().

There were a couple coding issues I cleaned up at the same time. E.g. if getSourceList() is empty it'll crash less. We should leave this empty if the original list is an array (and thus not a List we can lazy evaluate).

Strictly this field doesn't support array source, but it's a challenge to hard-reject arrays it without breaking semver.

### DIFF
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -216,7 +216,7 @@ class TagField extends DropdownField
             $this->setAttribute('multiple', 'multiple');
         }
 
-        if ($this->shouldLazyLoad) {
+        if ($this->getShouldLazyLoad()) {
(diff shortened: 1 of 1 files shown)


src/TagField.php
```diff
@@ -262,6 +267,20 @@ protected function getOptions()
         $ids = $values->column($this->getTitleField());
 
         $titleField = $this->getTitleField();
+        
+        if ($this->shouldLazyLoad) {
```
robbieaverill: Minor: can we use `$this->getShouldLazyLoad()`?

tractorcow: Done.

PR #120, opened 2018-10-10
Setup TagField to work within AssetAdmin (Fixes #107)

### LINKED ISSUE #107: TagField doesn't render in AssetAdmin
Trying to add a `TagField` to `Image` and get an empty dropdown and no ability to add new tags. I imagine that React is getting in the way - happy to take a crack at it but have no idea what direction to go in.

To reproduce:

```php
class GalleryImageExtension extends DataExtension
{
    private static $many_many = [
        'GalleryImageTags' => GalleryImageTag::class,
    ];
}
```

```php
class GalleryImageFormFactoryExtension extends Extension
{
    public function updateFormFields(FieldList $fields, $controller, $formName, $context)
    {
        $image = $context['Record'];
        $tagField = TagField::create(
            'GalleryImageTags',
            'Tags',
            GalleryImageTag::get(),
            $image->GalleryImageTags()
        );
        $fields->addFieldToTab('Editor.Details', $tagField);
    }
}
```

```php
class GalleryImageTag extends DataObject
{
    private static $db = [
        'Title' => 'Varchar(50)',
    ];

    private static $belongs_many_many = [
        'GalleryImages' => Image::class,
    ];

    private static $table_name = 'GalleryImageTag';
}
```

```yml
SilverStripe\Assets\Image:
  extensions:
    - 'GalleryImageExtension'
SilverStripe\AssetAdmin\Forms\ImageFormFactory:
  extensions:
    - 'GalleryImageFormFactoryExtension'
```

wilr:
Form field now appears, can be added and searched.

This also fixes missing label
screen shot 2018-10-10 at 4 05 55 pm


client/src/components/TagField.js
```diff
@@ -12,19 +13,24 @@ class TagField extends Component {
...
 
     this.onChange = this.onChange.bind(this);
+    this.handleOnBlur = this.handleOnBlur.bind(this);
     this.getOptions = this.getOptions.bind(this);
   }
 
   onChange(value) {
     this.setState({
-      value
+      value: value
```
robbieaverill: This change is functionally equivalent to what was already there with ES6

client/src/components/TagField.js
```diff
@@ -12,19 +13,24 @@ class TagField extends Component {
...
     });
 
     if (typeof this.props.onChange === 'function') {
       this.props.onChange(value);
     }
   }
 
+  handleOnBlur (e) {
+
+  }
```
robbieaverill: What's the purpose of this?

src/TagField.php
```diff
@@ -200,28 +202,27 @@ public function setSourceList($sourceList)
...
-            'value' => $this->Value(),
-            'disabled' => $this->isDisabled() || $this->isReadonly(),
-        ];
-        if (!$this->getShouldLazyLoad()) {
-            $schema['options'] = array_values($this->getOptions()->toNestedArray());
+        Requirements::css('silverstripe/tagfield:css/select2.min.css');
+        Requirements::css('silverstripe/tagfield:css/TagField.css');
+
+        Requirements::javascript('silverstripe/tagfield:js/select2.js');
+        Requirements::javascript('silverstripe/tagfield:js/TagField.js');
```
robbieaverill: This looks like it's undoing the use of the bundles, maybe going back to what we had in SS3 - is this deliberate?

src/TagField.php
```diff
@@ -200,28 +202,27 @@ public function setSourceList($sourceList)
...
+
+        $this->addExtraClass('ss-tag-field');
+
+        if ($this->getIsMultiple()) {
+            $this->setAttribute('multiple', 'multiple');
+        }
+
+        if ($this->shouldLazyLoad) {
```
robbieaverill: Can you use `$this->getShouldLazyLoad()`

client/src/components/TagField.js
wilr: Hey man, see my comments on the react slack channel. Was a bug with react-select https://github.com/JedWatson/react-select/issues/805

robbieaverill: Ok sweet, can you add a docblock with that issue link in it maybe?

robbieaverill:
I think this PR mostly looks good, but there's a couple of breaking changes that would need to be adjusted and polished. Also I've just made a PR at #122 which will cause a couple of conflicts with adding the same logic in this PR in place

wilr:
Happy if @robbieaverill you want to incorporate the bits needed into #122 and close this.

robbieaverill:
@wilr I might try and do it on bug day tomorrow

robbieaverill:
(Context: #122 with these changes applied)

Hrmmm, I've got the fields rendering correctly in asset admin but when I have `'name' => $this->getName() . '[]';` redux-form won't validate or store the tag values correctly, which means (A) it won't add the tag in the UI and (B) it won't save it (related issue: https://github.com/silverstripe/silverstripe-admin/issues/639)

If I use a regular `$this->getName()`, tags end up getting pushed into the DOM in multiple hidden inputs with the same name.

Does this PR actually save @wilr?

@ScopeyNZ are you able to provide any advice here in terms of how this field interacts with redux-form? Again related: https://github.com/silverstripe/silverstripe-admin/issues/639

wilr:
@robbieaverill Yep saves perfectly, been running in prod for a couple weeks now. But this is using composer.lock (https://gist.github.com/wilr/68ed4e6eb9f063da4e00dec6831e0a98). Note the `[]` was to copy how `CheckboxSetField` passed its' array.

robbieaverill:
@wilr In this PR I've noticed that I get an undefined index error when using the fields on a non-React form, and that StringTagField isn't working in the React context yet. It does look like this PR does make TagField work correctly in a React context, so thank you for that! I'll try and work out why #122 isn't based off of this and move it to that.

robbieaverill:
Yeah OK, I can get it working in _either_ entwine or React context but not both at the same time (this PR is the same). I've raised an issue at https://github.com/silverstripe/silverstripe-admin/issues/755 for tracking. I suspect we'll have to refactor more of the logic in TagField to get this to work, and probably do as you've suggested here and switch to extending MultiSelectField.

I'll leave this PR open

ScopeyNZ:
It took me wayyyy too long but I finally got around to having a go at fixing this. I've tested this using the blog module and using elemental in-line editing to render a tag field in the React context.

For the react context you'll need the admin PR merged as a dependency (https://github.com/silverstripe/silverstripe-admin/pull/912) otherwise the values won't appear in the field.

robbieaverill:
@ScopeyNZ is that the only dependency for this PR, e.g. once the admin PR is merged will this PR fix the original issue?

ScopeyNZ:
Yeah we spoke about this a little offline. The admin PR ensures that the field value is correctly set from redux state. So I _think_ this only matters in the elemental context.

ScopeyNZ:
Thanks @NightJar . There was some confusion about IDs vs label titles. Should be squared away now.

client/src/components/TagField.js
```diff
@@ -43,6 +43,49 @@ class TagField extends Component {
     return this.fetchOptions(input);
   }
 
+  /**
+   * Handle a change, either calling the change handler provided (if controlled) or updating
+   * internal state of this component
+   *
+   * @param value
```
robbieaverill: Needs a type

client/src/components/TagField.scss
```diff
@@ -1,5 +1,5 @@
-@import "~react-select/dist/react-select.css";
+@import "~react-select/scss/default";
 
-.ss-tag-field .Select--multi .Select-value {
+.Select--multi .Select-value {
```
robbieaverill: This change is a little concerning - looks like it'd be then applying to any react selects in the CMS...?

src/TagField.php
```diff
@@ -212,43 +216,29 @@ public function Field($properties = [])
...
 
         if (!$this->getShouldLazyLoad()) {
             $schema['options'] = array_values($this->getOptions()->toNestedArray());
         } else {
-            if ($this->Value()) {
-                $schema['value'] = $this->getOptions(true)->toNestedArray();
-            }
             $schema['optionUrl'] = $this->getSuggestURL();
         }
+        $this->setAttribute('data-schema', json_encode($schema));
```
robbieaverill: This is already in getAttributes, is that a double up?

NightJar:
Thanks @ScopeyNZ - can confirm latest update has resolved that issue :)

client/src/components/TagField.scss
ScopeyNZ: Yeah. I figured that'd be okay - I only did this because I had to work around this requirement with entwine and then again with storybook. I'll just change it back and workaround in storybook though.

ScopeyNZ:
Ok I've addressed feedback and rebased for a minor release. Thanks for the review @robbieaverill & @NightJar!

robbieaverill:
Good result everyone!

PR #122, opened 2018-11-15
FIX TagField and StringTagField now work correctly on SilverStripe 4.3

robbieaverill:
High level changes:

* TagField and doesn't return `tag` for its Type(), breaking the template layout in SS 4.3 (see https://github.com/silverstripe/silverstripe-admin/issues/747)
* Removed legacy css and javascript files that were no longer exposed, therefore not in use
* Update array syntax and PHPDocs in some classes
* Modify StringTagField so that it suggests options from the input array provided to it, rather than the values that you've already saved into the record that owns the field
* Update documentation

Resolves https://github.com/silverstripe/silverstripe-blog/issues/558


NightJar:
I see removed JS, but I don't see replacement. Does this mean that it relies on a react form field provided by `admin` (i.e. default for `DropdownField`)?

ScopeyNZ:
> I see removed JS, but I don't see replacement. Does this mean that it relies on a react form field provided by `admin` (i.e. default for `DropdownField`)?

The prior (incomplete) upgrade already added a react tagfield which is being used. It's all part of the bundle now 🙂 . I'm going to try and give this a good run through tomorrow! Thanks for the efforts.

NightJar:
Ah, nice :)

ScopeyNZ:
It works well. I think #561 is a killer though... I added a commit to address this.

Otherwise I'm happy now 🙂

robbieaverill:
I'm having a lot of trouble getting this to work in both an Entwine and a React context. I thought that the conversion would be quick so I'd tackle it as part of https://github.com/silverstripe/silverstripe-blog/issues/558, however I've proven myself wrong so I'm going to reset this PR back to fixing that bug and we will tackle the React form field issue separately (#107 and #120)

robbieaverill:
@ScopeyNZ sorry I've removed your commit, we can make a follow up PR afterwards

PR #124, opened 2018-12-04
FIX Adding a debounce to fetching lazy loaded tags with tag field

### LINKED ISSUE #123: Add a slight delay before performing lazy load searches
@robbieaverill commented on [Thu Nov 15 2018](https://github.com/silverstripe/silverstripe-blog/issues/561)

**Version: 2.2.x-dev** (React version)

When I type "test" into a lazy loadable TagField, I see four network requests:

image

We could put a half second resettable timeout in before we perform the XHR request to help prevent this.

### PR

#124

ScopeyNZ:
Resolves #123

This imports lodash's `debouce` function to defer loading tagfield suggestions until there is no input change for 500ms.


ScopeyNZ:
This was pushed as part of a previous PR but the Jest tests failed in a non-trivial way. They are still passing locally for me so I'll let CI guide me.

ScopeyNZ:
Ok I've updated the test as recommended by some comments in [this](https://github.com/facebook/jest/issues/3465) issue using a `setTimeout`.

🤞

Issue #74, opened 2016-02-15
[SS 3.3] Error from entwine


gregsmirnov:
Steps to reproduce:
1. Create PageTag class and add m-m relation to Page as described in Readme
2. Open CMS to edit existing page in Chrome browser (used v48)
3. Open Developer tools
4. Click Save button
5. Observe exception in the JS console

``` js
jquery.entwine-dist.js?m=1455273235:888 Uncaught exception TypeError: Cannot read property 'apply' of undefined(…) in onunmatch on
 <select name=​"Tags[]​" class=​"tag ss-tag-field select2-hidden-accessible" id=​"Form_EditForm_Tags" multiple style=​"display:​ inline-block;​" tabindex=​"-1" aria-hidden=​"true">​</select>​
jquery.entwine-dist.js?m=1455273235:891 Stack Trace:
TypeError: Cannot read property 'apply' of undefined
    at HTMLSelectElement.one [as f] (http://example.com/framework/thirdparty/jquery-entwine/dist/jquery.entwine-dist.js?m=1455273235:1011:37)
    at jQuerySub.$.extend._super (http://example.com/framework/thirdparty/jquery-entwine/dist/jquery.entwine-dist.js?m=1455273235:1153:13)
    at jQuerySub.$.entwine.onunmatch (http://example.com/framework/admin/javascript/LeftAndMain.js?m=1455273230:1286:10)
    at Array.proxy [as onunmatchproxy] (http://example.com/framework/thirdparty/jquery-entwine/dist/jquery.entwine-dist.js?m=1455273235:1867:23)
    at HTMLDocument.<anonymous> (http://example.com/framework/thirdparty/jquery-entwine/dist/jquery.entwine-dist.js?m=1455273235:2043:16)
    at HTMLDocument.jQuery.event.dispatch (http://example.com/framework/thirdparty/jquery/jquery.js?m=1455273235:3332:9)
    at HTMLDocument.elemData.handle.eventHandle (http://example.com/framework/thirdparty/jquery/jquery.js?m=1455273235:2941:28)
    at Object.jQuery.event.trigger (http://example.com/framework/thirdparty/jquery/jquery.js?m=1455273235:3210:12)
    at jQuery.fn.extend.triggerHandler (http://example.com/framework/thirdparty/jquery/jquery.js?m=1455273235:3874:24)
[...]

PR #128, opened 2018-12-11
Update faulty import of TagField component

svenV88:
#74 Should help fix the last mentioned error.

### DIFF
--- a/client/src/boot/registerComponents.js
+++ b/client/src/boot/registerComponents.js
@@ -1,5 +1,5 @@
 import Injector from 'lib/Injector';
-import TagField from 'components/TagField';
+import TagField from '../components/TagField';
 
 export default () => {
   Injector.component.registerMany({


ScopeyNZ:
Hey @svenV88 thanks for the patch. Did you notice any difference testing this? I only ask because it appears the code you have updated does not actually cause any difference in the bundled files. Normally when we make javascript changes you have to also commit bundled files (`client/dist`) but your PR does not include them. Normally I would ask you to make sure your bundled files are committed and pushed but we have a CI task that fails the build if the bundled files are not included:

https://github.com/silverstripe/silverstripe-tagfield/blob/master/.travis.yml#L37-L38

Because the builds are failing it appears that what you have is functionally equivalent - which is what I expected. In our modules we make `client/src` a base path for imports.

Anyway - if this fixed the issue for you please let me know. It would surprise me.

svenV88:
@ScopeyNZ I indeed didn't build the project as I thought it wouldn't be necessary with such a change and couldn't really come up with a clever way to test it in my current project. Now I realise this was a horrible decision (so thanks for not accepting it straight away). Will try to test the code this time and see if I can find anything.

robbieaverill:
I've pulled this in locally and run `yarn build`. It doesn't affect the client/dist output, which means there is no functional change from this pull request.

Having said that, I don't mind merging it anyway.

Issue #129, opened 2018-12-18
TagField does not render on initial page load


mrbookt:
SS 4.3
TagField 2.2.1

TagField does not render on initial page load in model admin. User have to refresh page to see tag field.

javascript warning:

> Component TagField does not exist

![ss4_3_tagfield_2_2_1](https://user-images.githubusercontent.com/839784/50145559-2afa6d00-02b2-11e9-9b7f-ef6d832287b5.png)

after refresh:
![ss4_3_tagfield_2_2_1_refresh](https://user-images.githubusercontent.com/839784/50145557-2a61d680-02b2-11e9-97e7-2295312f1c1e.png)

robbieaverill:
Hi @mrbookt, are you clicking into the ModelAdmin in order to load it the first time, or accessing the URL directly in your browser?

It sounds like the frontend assets are not loaded when the PJAX request is made (from clicking into a different area in the admin)

mrbookt:
Hi @robbieaverill if I access url directly  (and just replace record id to edit) everything works ok. While I am clicking ModelAdmin -> gridfield -> record edit TagField component is not loaded, I have to hit refresh button to load it

robbieaverill:
Ok cool, the fix is to add the CSS and JS requirements to LeftAndMain via YAML config, rather than add them in the FormField constructor. Here's an example: 

https://github.com/dnadesign/silverstripe-elemental/blob/147a6877c3c215258ed1934f55ea27b0529c2da1/_config/config.yml#L5-L9

Would you like to have a crack at making a pull request to fix this?

mrbookt:
Thanks, I'll make a pr soon

thats4shaw:
Ran into this today myself so i've made a pull request.

mrbookt:
@thats4shaw thanks

ScopeyNZ:
Closed in #131

ntd:
Any chance of having a release with the above fix included? Not being able to use it (out of the box) in the back end is IMO a quite big flow.

robbieaverill:
@ntd done: https://github.com/silverstripe/silverstripe-tagfield/releases/tag/2.2.2

PR #131, opened 2018-12-19
Add css and js to left and main

thats4shaw:
* Add yaml file
* Remove requirement calls from field constructors

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -1,0 +1,9 @@
+---
+Name: tagfield
+---
+
+SilverStripe\Admin\LeftAndMain:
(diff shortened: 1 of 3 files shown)


thats4shaw:
Fixes https://github.com/silverstripe/silverstripe-tagfield/issues/129.

robbieaverill:
We'd need to be careful - technically renaming the config block could be a breaking change if people are doing their own config `After: tagfield`

ScopeyNZ:
It's a new config block though?

robbieaverill:
Right - I only looked at the diff

Issue #132, opened 2019-01-23
TagField does not show matching tags after searching


Kethatril:
When searching for a tag using a TagField that is lazy loading, the matching results don't show and the loading spinner doesn't stop until you either continue typing, backspace or leave the tagfield and re-search.

![initial-search](https://user-images.githubusercontent.com/2230148/51637880-c5d91c00-1fc1-11e9-9386-cfd67f049822.png)

![after](https://user-images.githubusercontent.com/2230148/51637912-d8535580-1fc1-11e9-94fb-d0cffcf8ee9d.png)

Kethatril:
https://github.com/JedWatson/react-select/issues/3075 Is an explanation of why this is happening, I'll looking at making a fix.

Kethatril:
Easiest fix would be to replace lodash/debounce with a promise based debounce like https://github.com/bjoerge/debounce-promise.

PR #133, opened 2019-01-24
Replace lodash.debounce with a promise based debounce

Kethatril:
Fixes https://github.com/silverstripe/silverstripe-tagfield/issues/132


ScopeyNZ:
Nice. I had a play with the old version. It was really buggy.

New version works perfectly. Thanks heaps for the effort @Kethatril.

PR #137, opened 2019-02-19
FIX StringTagField now works with SS-2018-021/CVE-2019-5715 by serialising arrays before write

robbieaverill:


ScopeyNZ:
Ran CI locally and it's green. I think it's prudent to merge before CI finishes in this case. All our travis workers are blocked for the next few hours anyway.

PR #142, opened 2019-04-17
fixed the issue with filters being taken off from rendering values

fonsekaean:


src/TagField.php
```diff
@@ -275,7 +275,7 @@ protected function getOptions($onlySelected = false)
 
         if ($values) {
             if (is_array($values)) {
-                $values = DataList::create($dataClass)->filter($titleField, $values);
+                $values = $source->filter($titleField, $values); // DataList::create($dataClass)->filter($titleField, $values);
```
robbieaverill: Seems sensible. Would you mind removing the commented out code here, and updating the phpcs violation on line 393?

fonsekaean: @robbieaverill fixed those.

src/TagField.php
```diff
@@ -390,6 +390,9 @@ protected function getOrCreateTag($term)
             $record = Injector::inst()->create($dataClass);
             $record->{$titleField} = $term;
             $record->write();
+            if ($source instanceof SS_List) {
+                $source->add($record);
+            }
```
robbieaverill: Would you mind explaining what this change does? As far as I can tell it's not used, since `$record` is what is returned

fonsekaean: This adds the new tag created to the Many_Many / Has_many list. So it completes the relation.

robbieaverill: Oh of course, because `->add()` internally writes the `$record`

robbieaverill:
Thanks for the patch!

PR #156, opened 2020-02-06
Add custom readonly field for StringTagField

mlewis-everley:
If I try to view `StringTagField` as a `ReadonlyField`, I get the warning:

```
[Warning] array_key_exists(): The first argument should be either a string or an integer
GET /site/public/admin/messages/App-Model-Message/EditForm/field/App-Model-Message/item/3/view

Line 32 in /path/to/site/vendor/silverstripe/framework/src/Forms/SingleLookupField.php
Source

23     /**
24      * @return mixed|null
25      */
26     protected function valueToLabel()
27     {
28         $value = $this->value;
29         $source = $this->getSource();
30         $source = ($source instanceof Map) ? $source->toArray() : $source;
31 
32         if (array_key_exists($value, $source)) {
33             return $source[$value];
34         }
35 
36         return null;
37     }
38 
```

This is because `SingleLookupField` expects value to be a string or int, not an array.

This PR adds an alternate `ReadOnlyField`, that renders the value into the `ReadonlyField` as a comma separated string (as `ReadonlyTagField`). 

## Issue
- https://github.com/silverstripe/silverstripe-tagfield/issues/232

### DIFF
--- a/tests/StringTagFieldTest.php
+++ b/tests/StringTagFieldTest.php
@@ -133,5 +133,22 @@ class StringTagFieldTest extends SapphireTest
 
         $readOnlyField = $field->performReadonlyTransformation();
         $this->assertEquals(ReadonlyStringTagField::class, get_class($readOnlyField));
+        $this->assertEquals('', $readOnlyField->Value());
+
+        $field_two = new StringTagField('Tags');
+        $field_two->setSource(['Test1', 'Test2', 'Test3']);
+
+        $field_two->setValue(['Test1', 'Test2']);
+        $field_two_readonly = $field_two->performReadonlyTransformation();
+        $this->assertEquals('Test1, Test2', $field_two_readonly->getFieldValue());
+
+        // Ensure an invalid value isn't rendered
+        $field_two->setValue(['Test', 'Test1']);
+        $field_two_readonly = $field_two->performReadonlyTransformation();
+        $this->assertEquals('Test1', $field_two_readonly->getFieldValue());
+
+        $field_two->setValue(['Test2']);
+        $field_two_readonly = $field_two->performReadonlyTransformation();
+        $this->assertEquals('Test2', $field_two_readonly->getFieldValue());
     }
 }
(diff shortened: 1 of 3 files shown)


mlewis-everley:
I have added this to 2.1.* as this is the branch I am using and getting the error. I tried upgrading but the issue also exists in the latest version, so hopefully it van be up merged...

mlewis-everley:
I have no idea why travis is failing this one, looks like a dependency issue. I have run the tests locally and they all run OK.

tests/StringTagFieldTest.php
```diff
@@ -120,4 +121,17 @@ protected function getNewRequest(array $parameters)
...
+     * Test read only fields are returned
+     */
+    public function testReadonlyTransformation()
+    {
+        $record = $this->getNewStringTagFieldTestBlogPost('BlogPost2');
+        $field = new StringTagField('Tags');
+        $field->setRecord($record);
+
+        $readOnlyField = $field->performReadonlyTransformation();
+        $this->assertEquals(ReadonlyStringTagField::class, get_class($readOnlyField));
```
ScopeyNZ: You could do an assertion here on the tag string too?

ScopeyNZ (commented):
Yeah, `2.1` is a pretty old branch. The dev versions of `4.0.x` have been deleted from GitHub. I'll leave it to one of SS's OS team to fix that composer.json.

tests/StringTagFieldTest.php
mlewis-everley: Right, I think I have added this as you wanted.

ScopeyNZ: Nice... Did that show you a bug by chance?

mlewis-everley: Not that I recall, is there anything in particular I should be looking for?

ScopeyNZ: Haha nope, never-mind. I just noticed that the revision that added more tests also slightly tweaked the implementation and was curious if the updated test helped you find bugs. I always love it when that happens 😅 .

I don't work too closely with SS right now, so I'm not really in the position to check this out locally and play around, or fix the `composer.json`. Hopefully one of the SSLtd team or core team will get around to it soon!

GuySartorelli:
Closed in favour of https://github.com/silverstripe/silverstripe-tagfield/pull/233

PR #162, opened 2020-07-06
BUG Fix field not working in react (asset-admin / elemental)

### LINKED ISSUE #150: Tags don't display in Assetadmin
Hi together,

when creating a TagFiled 2.4 in an extension of assets, the tags are not visible after saving.
the tags are written in the database, but not displayed.

```php
class MosaikAssetFormFactoryExtension extends Extension {
   public function updateFormFields(FieldList $fields, $controller, $formName, $context)
    {
        if (isset($context['Record'])) {
           $MFile = $context['Record'];
           $fields->insertAfter('Title', TagField::create('Keywords', 'Schlagworte', MosaikKeywords::get(), 
              $MFile->Keywords(), 'Keyword')->setShouldLazyLoad(true)->setCanCreate(true)- 
                >setIsMultiple(true));
          // .....
          }
      }
}
```

in a DataObject class TagField works perfect.

thanks for your answer

Klausjulius

tractorcow:
Fixes #150

The issue is to do with the way that react serialises selected values.

I've actually tested this in elemental module, which uses react forms, but it should also fix asset-admin fields.

### DIFF
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -6,7 +6,6 @@ use Exception;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
-use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Forms\MultiSelectField;
 use SilverStripe\Forms\Validator;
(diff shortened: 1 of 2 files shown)


ScopeyNZ:
🤯 . I feel like this issue has been raised in so many forms. 

Are you sure that the method is not called with an array of values? As opposed to an array of keyed arrays?

tractorcow:
> Are you sure that the method is not called with an array of values? As opposed to an array of keyed arrays?

It normally is; If you use this field in a normal entwine form, it posts as a basic array of values. Hence the `else` part of the ternary at `                $values[] = isset($item['Value']) ? $item['Value'] : $item;` ($item is the value).

The `Value` key is only set, seemingly, when the data comes from react. It seems safer to not make any assumptions about the data source and simply support both formats (as covered by unit tests).

ScopeyNZ:
Haha right. Perhaps I shouldn't review stuff first thing in the morning.

ScopeyNZ (approved):

Issue #171, opened 2020-09-21
Multiple Blogs with same category names, the tag field will display the selected category multiple times


obj63mc:
Affects SS 4.x, TagField Latest

First not sure if this is a Tag Field issue or a https://github.com/silverstripe/silverstripe-blog or more of a general SilverStripe issue with form scaffolding

To reproduce - 
1. Create a new SilverStripe Site
2. Install the Tagfield and Blog Module
3. Add two blogs
4. On Blog One add Categories 'Foo', 'Bar', 'Baz'
5. On Blog Two add Categories 'Foo' and 'Bar'
6. On Blog One - add a new post with the category 'Foo' - when searching the form will show only one option and when added to the field it is listed once.
7. Save the Page
8. On refresh the form field will display 'Foo' twice
9. Add Category 'Baz' to the post
10. Save Page will show - 'Foo, Foo, Baz'

I have some screenshots attached below. As with the testing in steps 9 and 10 if the category name isn't duplicated anywhere it shows singularly as it should. When looking in the database all is correct there, and when looking at the post counts and dumps of the posts categories the count is correct so not sure why the response back from the server is duplicated after the page is saved.

## Blog One with categories defined - 

example test is the Careers Category
![1BlogOne](https://user-images.githubusercontent.com/443120/93826959-f1485a00-fc2d-11ea-8e73-b3521536ee82.png)

## Blog Two with the same categories defined - 

you can see that category two shows proper count no posts in Careers
![2BlogTwo](https://user-images.githubusercontent.com/443120/93826958-f0afc380-fc2d-11ea-9a85-af6895ab07d1.png)

## Before Save adding the Category - Page ID 43, Category ID 7
![3BeforeSave](https://user-images.githubusercontent.com/443120/93826956-f0172d00-fc2d-11ea-8327-a40f8881d794.png)

## After Save - category is showing twice.
![4AfterSave](https://user-images.githubusercontent.com/443120/93826954-ef7e9680-fc2d-11ea-8704-5777a5921d6c.png)

## Database Table of Blog Categories - table name BlogCategory
![Database](https://user-images.githubusercontent.com/443120/93827341-b135a700-fc2e-11ea-9f7f-d21cbc6c24e0.png)

## Database table of saved relations - table name BlogPost_Categories

shows that data isn't duplicated it is singular per post as expected. only one row for Blog Post ID 43 and Category ID 7
[...]

PR #173, opened 2020-10-19
Fix for #171 - Issue when utilizing multiple of same parent class

obj63mc:
This fix should take care of https://github.com/silverstripe/silverstripe-tagfield/issues/171

Essentially checking first if the source is a datalist of some sort, if so utilize it instead of creating a new one.

### DIFF
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -276,8 +276,12 @@ class TagField extends MultiSelectField
         // Convert an array of values into a datalist of options
         if (!$values instanceof SS_List) {
             if (is_array($values) && !empty($values)) {
-                $values = DataList::create($dataClass)
-                    ->filter($this->getTitleField(), $values);
+                if(is_subclass_of($source, DataList::class) || is_a($source, DataList::class)){
+                    $values = $source->filter($this->getTitleField(), $values);
+                } else {
+                    $values = DataList::create($dataClass)
+                        ->filter($this->getTitleField(), $values);
+                }
             } else {
                 $values = ArrayList::create();
             }
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -276,7 +276,7 @@ class TagField extends MultiSelectField
         // Convert an array of values into a datalist of options
         if (!$values instanceof SS_List) {
             if (is_array($values) && !empty($values)) {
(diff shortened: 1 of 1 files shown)


src/TagField.php
```diff
@@ -276,8 +276,12 @@ protected function getOptions($onlySelected = false)
         // Convert an array of values into a datalist of options
         if (!$values instanceof SS_List) {
             if (is_array($values) && !empty($values)) {
-                $values = DataList::create($dataClass)
-                    ->filter($this->getTitleField(), $values);
+                if (is_subclass_of($source, DataList::class) || is_a($source, DataList::class)) {
```
micschk: @obj63mc `is_a($source, DataList::class)` Checks if `$source` is of class `DataList` **OR** has `DataList` as one of its parents. So I think `is_subclass_of($source, DataList::class)` can be omitted.

micschk: @obj63mc `is_a($source, DataList::class)` Checks if `$source` is of class `DataList` **OR** has `DataList` as one of its parents. So I think `is_subclass_of($source, DataList::class)` can be omitted.

micschk (changes requested):
I can confirm this PR fixes the issue, good catch!
I did notice one redundant check which may be left out.

micschk:
@maxime-rainville I can confirm this PR fixes the issue.
Would be nice if this could be merged & new tag created?
I did find one redundant check, which may be left out.

@obj63mc `is_a($source, DataList::class)` Checks if `$source` is of class `DataList` OR has this class as one of its parents. So I think the [first check]() (`is_subclass_of($source, DataList::class)`) can be omitted.

src/TagField.php
obj63mc: Check that in so should be good to go now

obj63mc:
I checked in an update for this, wasn't 100% sure whether is_a would include subclasses or not so should be good now

mooror:
I have come across this issue as well. Any timeline on the merging in of this PR?

michalkleiner (approved):
Would be good to add tests for the updated mechanism but not critical as to keep blocking the PR any longer.

PR #174, opened 2020-11-10
MNT Use shared travis config, use sminnee/phpunit

emteknetnz:


emteknetnz:
@dnsl48 The travis failure is because the js/css for this was built along time ago with node 6 instead of node 10, so the NPM_TEST is failing because of that https://travis-ci.com/github/silverstripe/silverstripe-tagfield/jobs/431656163

This does not affect that end-user functionality of this, it only makes future development harder

We should merge this PR as is for now and fix node 6 at a later date

dnsl48 (approved):

PR #185, opened 2021-06-18
Fix StringTagField initialisation (fixes #148)

### LINKED ISSUE #148: StringTagField not initialized via Entwine
Since ["FIX Ensure tagfield is compatible with both React and Entwine contexts"](https://github.com/silverstripe/silverstripe-tagfield/commit/e97a82f0b7d7484a16e783b9432847659f8b0a9d) the StringTagField is no longer initialized via Entwine as it does not have the class "entwine" attached, which is required due to the [change in initialization](https://github.com/silverstripe/silverstripe-tagfield/commit/e97a82f0b7d7484a16e783b9432847659f8b0a9d#diff-5f55185ac07dc2877fd7f16a7cec6d4c). I assume that [adding the entwine class](https://github.com/silverstripe/silverstripe-tagfield/commit/e97a82f0b7d7484a16e783b9432847659f8b0a9d#diff-d072a5c333026c98dcf417dafdc8467dL204) in the Field method should be sufficient.

lozcalver:


dhensby (commented):
Should we retain the existing class?

lozcalver:
I guess we could do - it was removed in the original PR though, so I guess it’s not used for anything https://github.com/silverstripe/silverstripe-tagfield/commit/e97a82f0b7d7484a16e783b9432847659f8b0a9d#diff-5d6ff8240577bf0a3073f293f0d8d5ea9d096e50ce2f75ca65f6e17aa682a8a5L204

Issue #153, opened 2019-11-14
issue when using Taxonomies, need a error suggestion/ add to docs


vinstah:
When using a simple $many_many to TaxonomyTerm I get the error 

[Emergency] Uncaught SilverStripe\ORM\Connect\DatabaseException: Couldn't run query: SELECT AS "Title", "TaxonomyTerm"."Sort" AS "_SortColumn0" FROM "TaxonomyTerm" INNER JOIN "ProductPage_Tags" ON "ProductPage_Tags"."TaxonomyTermID" = "TaxonomyTerm"."ID" WHERE ("ProductPage_Tags"."ProductPageID" = ?) ORDER BY "_SortColumn0" ASC
```

private static $many_many =  [
     'Tags'    => TaxonomyTerm::class,
];

TagField::create(
  'Tags',
  'Tags',
  TaxonomyTerm::get(),
  $this->Tags()
)
```

The problem isn't immediately obvious as the TagField is looking for TaxonomyTerm.Title which doesn't exist, I had to hunt through the code to find the constructor had an extra param for title field

```
public function __construct($name, $title = '', $source = [], $value = null, $titleField = 'Title')
```

### PRs
- https://github.com/silverstripe/silverstripe-tagfield/pull/201

emteknetnz:
Hi, thanks for submitting this

Could you please clarify the error you're getting.  It seems as though the SQL you've supplied above is missing the error that MySQL is throwing e.g. "Missing column xyz"

Also, if you had any ideas around a proposed solution, that would also be helpful

emteknetnz:
Note quite sure why this closed

vinstah:
I only got notified today about this missed the comment for 3 years,

looks like @wilr closed this since they updated the docs which would prevent the error by using `->setTitleField('Name');`

wilr:
@emteknetnz Closed because its fixed in my commit

emteknetnz:
This [PR](https://github.com/silverstripe/silverstripe-tagfield/pull/201) hasn't been merged yet

emteknetnz:
Linked PR has been merged

PR #201, opened 2021-12-18
Selected tag is not shown when isMutliple is false

### LINKED ISSUE #195: Selected tag is not shown when isMutliple is false
I have no clue if I am using it wrong, but if `->setIsMultiple(false)` is set a selected tag title is not displayed.
But the record in the many_many junction table is created and the `X` also shows that some relation is present.

Screenshot

### PRs
- https://github.com/silverstripe/silverstripe-tagfield/pull/201

wilr:
Fixes #195, #153


client/src/components/TagField.js
```diff
@@ -128,12 +123,28 @@ class TagField extends Component {
...
+            Object.keys(passThroughAttributes.value)[0]
+          ];
+
+        if (typeof value === "object") {
+          passThroughAttributes.value = value;
+        }
+      }
+    }
+
+    console.log(this.props);
```
emteknetnz:
```suggestion
```

src/TagField.php
```diff
@@ -304,36 +319,10 @@ protected function getOptions($onlySelected = false)
...
 
         $source->each($addOption);
+
         return $options;
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function setValue($value, $source = null)
```
emteknetnz, line 313: This is backwards incompatible API change, we cannot remove this public function

src/TagField.php
```diff
@@ -368,80 +358,125 @@ public function setSource($source)
...
+
+        if (is_array($values)) {
+            return $values;
+        }
+
+        if ($values instanceof SS_List) {
+            return $values->column($this->getTitleField());
+        }
+
+        return [trim($values)];
```
emteknetnz:
```suggestion
        return [trim((string) $values)];
```
?

src/TagField.php
```diff
@@ -368,80 +358,125 @@ public function setSource($source)
...
-                'name' => $this->getName() . '[]',
+                'name' => $name,
                 'style' => 'width: 100%',
                 'data-schema' => json_encode($this->getSchemaData()),
             ]
         );
     }
 
+
+    protected function getListValues($values)
```
emteknetnz:
```suggestion
    protected function getListValues($values): array
```

client/src/components/TagField.js
```diff
@@ -1,10 +1,10 @@
-import React, { Component } from 'react';
-import Select from 'react-select';
-import fetch from 'isomorphic-fetch';
-import fieldHolder from 'components/FieldHolder/FieldHolder';
-import url from 'url';
-import debounce from 'debounce-promise';
-import PropTypes from 'prop-types';
+import React, { Component } from "react";
```
emteknetnz: There's a lot of JS changes that look like your IDE may have done some automated formatting.  Double quotes are not standard in this codebase, so these change will need to be reverted. It's causing linting errors in travis - https://app.travis-ci.com/github/silverstripe/silverstripe-tagfield/jobs/552813829

readme.md
```diff
@@ -2,23 +2,27 @@
 
 Custom tag input field, for SilverStripe.
 
-[![Build Status](https://api.travis-ci.com/silverstripe/silverstripe-tagfield.svg?branch=2)](https://travis-ci.com/silverstripe/silverstripe-tagfield)
-[![Code Quality](http://img.shields.io/scrutinizer/g/silverstripe/silverstripe-tagfield.svg?style=flat)](https://scrutinizer-ci.com/g/silverstripe/silverstripe-tagfield)
-[![Code coverage](https://codecov.io/gh/silverstripe/silverstripe-tagfield/branch/master/graph/badge.svg)](https://codecov.io/gh/silverstripe/silverstripe-tagfield)
+[![Build
```
emteknetnz: These long lines weren't failing any existing linting rules, and now won't match other modules. Please revert

src/TagField.php
```diff
@@ -567,4 +600,14 @@ public function getSchemaStateDefaults()
 
         return $data;
     }
+
+
+    public function getSchemaDataType()
```
emteknetnz:
```suggestion
    public function getSchemaDataType(): string
```

src/TagField.php
```diff
@@ -368,80 +358,125 @@ public function setSource($source)
...
-     * @param  string $term
+     * @param string $value
+     *
      * @return DataObject|bool
      */
-    protected function getOrCreateTag($term)
+    protected function getOrCreateTag($value)
     {
+        if (is_array($value)) {
+            $value = isset($value['Value']) ? $value['Value'] : '';
```
emteknetnz:
```suggestion
            $value = $value['Value'] ?? '';
```

src/TagField.php
```diff
@@ -567,4 +600,14 @@ public function getSchemaStateDefaults()
...
     }
+
+
+    public function getSchemaDataType()
+    {
+        if ($this->getIsMultiple()) {
+            return self::SCHEMA_DATA_TYPE_MULTISELECT;
+        } else {
+            return self::SCHEMA_DATA_TYPE_SINGLESELECT;
+        }
```
emteknetnz:
```suggestion
        if ($this->getIsMultiple()) {
            return self::SCHEMA_DATA_TYPE_MULTISELECT;
        }
        return self::SCHEMA_DATA_TYPE_SINGLESELECT;
```

emteknetnz (changes requested):
Looks good so far, I've left feedback and there's also a travis unit test failure to sort out too. I haven't manually replicated issue or tested PR yet.

src/TagField.php
wilr: @emteknetnz it's not removing it from the API, it's defined on the parent as public anyway.

client/src/components/TagField.js
wilr: Ok, single quotes wasn't configured in the `.editorconfig` I've added that preference now.

.editorconfig
```diff
@@ -13,6 +13,7 @@ trim_trailing_whitespace = true
 [{*.yml,package.json,*.js,*.scss}]
 indent_size = 2
 indent_style = space
+quote_type = single
```
wilr, line 16: @emteknetnz Added the preference for single quotes

wilr:
@emteknetnz reverted readme change and rebased.

src/TagField.php
emteknetnz: So why are we removing it here?  Would the parent method handle different inputs of $value and $source correctly?

Parent method `MultiSelectField::setValue()` is this for reference:
```php
    public function setValue($value, $obj = null)
    {
        // If we're not passed a value directly,
        // we can look for it in a relation method on the object passed as a second arg
        if ($obj instanceof DataObject) {
            $this->loadFrom($obj);
        } else {
            parent::setValue($value);
        }
        return $this;
    }
```

tests/StringTagFieldTest.php
```diff
@@ -171,6 +171,33 @@ public function testSchemaIsAddedToAttributes()
         $this->assertNotEmpty($attributes['data-schema']);
     }
 
+    /**
+     * @param array $expected
+     * @param TagFieldTestBlogPost $record
+     */
+    protected function compareExpectedAndActualTags(array $expected, TagFieldTestBlogPost $record)
```
emteknetnz: Why is method here?  It's not called within StringTagFieldTest?

emteknetnz (changes requested):
This does not seem to fix the original issue?  Setting `->isMultiple(false)` for both `TagField` and `StringTagfield` meant the tags did not show in the field

Code I used for testing below:

```php
<?php
use SilverStripe\TagField\TagField;
use SilverStripe\CMS\Model\SiteTree;
use SilverStripe\TagField\StringTagField;

class BlogPost extends SiteTree
{
    private static $many_many = [
        'BlogTags' => BlogTag::class
    ];

    private static $db = [
        'StringTags' => 'Text',
    ];

    public function getCMSFields()
    {
        /** @var TagField field */
        $field = TagField::create(
            'BlogTags',
            'Blog Tags',
            BlogTag::get(),
            $this->BlogTags()
        )
            ->setShouldLazyLoad(true)
            ->setCanCreate(true)
            ->setIsMultiple(true)
            ;

        $fields->insertAfter('Title', $field);

        $field = StringTagField::create(
            'StringTags',
            'StringTags',
            ['one', 'two'],
            explode(',', $this->Tags)
        )
            ->setShouldLazyLoad(true)
            ->setIsMultiple(true)
            ;
        $fields->insertAfter('Title', $field);

        return $fields;
    }
}

<?php

use SilverStripe\ORM\DataObject;

class BlogTag extends DataObject
{
    private static $db = [
        'Title' => 'Varchar(200)',
    ];

    private static $belongs_many_many = [
        'BlogPosts' => BlogPost::class
    ];
}

<?php

use SilverStripe\Admin\ModelAdmin;

class BlogTagAdmin extends ModelAdmin
{
    private static $url_segment = 'blog-tags';

    private static $managed_models = [
        BlogTag::class
    ];
}
```

wilr:
@emteknetnz sorry when rebasing I didn't rebuild the bundle.js so the bug would still be present. If you try that on your checkout it should be green. Just checking it out now.

tests/StringTagFieldTest.php
wilr: Removed method.

wilr:
@emteknetnz Updated the `bundle.js` correctly now after my re-base and confirmed that example works as intended (admin:1.10.2)

emteknetnz (changes requested):
Thanks, have tested locally, now behaves as expected

There's a failing unit test that will need to be resolved

wilr:
Failing test in Postgres was because of the fallback to `ID`. It didn't like doing `ID IN ("Test1")`.

I've removed the ID fallback now.

emteknetnz (changes requested):
That last commit broken functionality - I manually tested that regular `TagField` no longer saves correctly when adding `->setIsMultiple(true)`.  Reverting the last commit fixes this issue, though obviously it means the postgres unit test will also fail.

wilr:
@emteknetnz fixed that regression now.

emteknetnz (approved):
Tested locally, works good

PR #208, opened 2022-04-01
Make the write call of StringTagField::saveInto method configurable

dkliemsch:
With that PR its possible to opt-out of the immediate write call within the `saveInto` method of the `StringTagField` using a custom config like this:

```yml
---
Name: app-tagfieldconfig
After:
  - tagfieldconfig
---
SilverStripe\TagField\StringTagField:
  immediate_write_enabled: false
```

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -7,9 +7,3 @@ SilverStripe\Admin\LeftAndMain:
     - 'silverstripe/tagfield:client/dist/js/bundle.js'
   extra_requirements_css:
     - 'silverstripe/tagfield:client/dist/styles/bundle.css'
-
----
-Name: tagfieldconfig
----
-SilverStripe\TagField\StringTagField:
-  immediate_write_enabled: true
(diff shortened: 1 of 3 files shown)


dhensby:
I think we certainly need a test for this.

_config/config.yml
```diff
@@ -7,3 +7,9 @@ SilverStripe\Admin\LeftAndMain:
     - 'silverstripe/tagfield:client/dist/js/bundle.js'
   extra_requirements_css:
     - 'silverstripe/tagfield:client/dist/styles/bundle.css'
+
+---
+Name: tagfieldconfig
+---
+SilverStripe\TagField\StringTagField:
+  immediate_write_enabled: true
```
ChloeSartorelli: Instead of setting this in a yml file, please set it in the class itself and add a [deprecation notice](https://docs.silverstripe.org/en/4/contributing/release_process/#deprecation)

GuySartorelli (changes requested):

_config/config.yml
dkliemsch: Ok, PR updated. I guess it's ok to deprecate it for the next major version 3.0.0?

dkliemsch:
> I think we certainly need a test for this.

I've added two tests with the option enabled/disabled, hope that's ok?

michalkleiner (approved):

GuySartorelli (approved):

PR #213, opened 2022-06-27
MNT Use GitHub Actions CI

emteknetnz:
Issue https://github.com/silverstripe/gha-ci/issues/11

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,16 @@
+name: CI
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


GuySartorelli:
Noting for the sake of visibility:
Note that the CI didn't run because there were no runners available, and the workflow isn't on the repo yet so it wasn't queued

The fact that travis didn't run is a little concerning - but ultimately benign for this PR since nothing here will affect travis.

GuySartorelli (approved):

PR #214, opened 2022-06-28
fix: Deal with DataObjects or Int values

bummzack:
This PR fixes an issue with `has_one` relations and the TagField. Creating and updating works, but the value is not shown in the Field. I tested with #195, but this didn't solve the issue.

With this patch it's possible to use the `TagField` with `isMultiple = false` and `has_one` relations. Tested with Pages as well as inline editing via Gridfieldextensions.

### DIFF
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -402,6 +402,14 @@ class TagField extends MultiSelectField
             return $values->column($this->getTitleField());
         }
 
+        if ($values instanceof DataObject && $values->exists()) {
+            return [$values->ID];
+        }
+
+        if (is_int($values)) {
+            return [$values];
+        }
+
         return [trim((string) $values)];
     }
 
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -403,7 +403,7 @@ class TagField extends MultiSelectField
         }
 
         if ($values instanceof DataObject && $values->exists()) {
-            return [$values->{$this->getTitleField()}];
+            return [$values->{$this->getTitleField()} ?? $values->ID];
         }
 
         if (is_int($values)) {
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -403,7 +403,7 @@ class TagField extends MultiSelectField
         }
 
         if ($values instanceof DataObject && $values->exists()) {
-            return [$values->getField($this->getTitleField())];
+            return [$values->{$this->getTitleField()}];
         }
 
         if (is_int($values)) {
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -403,7 +403,7 @@ class TagField extends MultiSelectField
         }
 
         if ($values instanceof DataObject && $values->exists()) {
-            return [$values->ID];
+            return [$values->getField($this->getTitleField())];
         }
 
         if (is_int($values)) {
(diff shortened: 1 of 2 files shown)


src/TagField.php
```diff
@@ -402,6 +402,14 @@ protected function getListValues($values): array
             return $values->column($this->getTitleField());
         }
 
+        if ($values instanceof DataObject && $values->exists()) {
```
emteknetnz, line 405: We use `$this->getTitleField()` above, how come we're using `->ID` here?

emteknetnz (dismissed):

src/TagField.php
bummzack: Since either IDs or titles work (https://github.com/silverstripe/silverstripe-tagfield/blob/2/src/TagField.php#L289) I opted for ID since it's unambiguous and most likely more efficient.
Should I change it to `return [$values->getField($this->getTitleField())];` ?

emteknetnz: Makes sense for it to be consistent.

Is there a scenario where $values won't have a column matching the title field? Do we need a fallback to the ID field, or is title field always safe?

bummzack: @emteknetnz I changed it to use the title-field instead of the ID.

src/TagField.php
```diff
@@ -402,6 +402,14 @@ protected function getListValues($values): array
...
+            return [$values->getField($this->getTitleField())];
```
ChloeSartorelli: `getField()` will skip any `getTitle()` method that might be intended to override the db value.
```suggestion
            return [$values->{$this->getTitleField()}];
```

GuySartorelli (changes requested):
I think the concern that there may not be a title field is valid - this should fallback on the ID if no title field is available. Need a `$values->hasField()` check.

src/TagField.php
bummzack: Thanks for the review. Changed as suggested

ChloeSartorelli: Awesome, thanks for that. Can you please address the ID fallback as well?

bummzack: @GuySartorelli done

ChloeSartorelli: Thanks!

GuySartorelli (approved):

GuySartorelli:
I still can't get the value to display in the field - can you please include some instructions for how to set this up?
I've tried both of these:
- `TagField::create('BlogTag', 'Blog Tag', BlogTag::get(), $this->BlogTag())->setIsMultiple(false)`
- `TagField::create('BlogTag', 'Blog Tag', BlogTag::get(), $this->BlogTagID)->setIsMultiple(false)`

with the `has_one` as so:
```php
private static $has_one = [
    'BlogTag' => BlogTag::class
];
```

bummzack:
Hi @GuySartorelli 
The field name should have an `ID` suffix. Eg.
```
TagField::create('BlogTagID', 'Blog Tag', BlogTag::get())->setIsMultiple(false)
```

GuySartorelli:
Ahh okay. That makes sense. Can you please add that to the readme? Create a new section below "Relational Tags"

bummzack:
@GuySartorelli done

GuySartorelli (approved):
Thanks! Works well.

PR #220, opened 2022-11-07
FIX Ignore hidden inputs for tracking unsaved changes

### LINKED ISSUE #125: Changes should not be tracked until a tag is added
### Description

Currently when using tag field in the context of a CMS edit page the input is being marked as changed while searching for tags to select. Ideally a change would only be tracked when tags are added or removed (ie. different from the tags that were there originally)

### Steps to reproduce

- Add a tag field to CMS form for a page (or just edit a blog page for an example)
- Type in the tag field to search for tags
- Note that the save/publish buttons in the "south toolbar" show as green implying there are unsaved changes

renskorswagen:
Ignoring the hidden inputs in the React Select component will fix tracking unsaved changes.

Fixes #125


michalkleiner (approved):

michalkleiner:
Thanks @renskorswagen, released as 2.10.1

PR #227, opened 2023-01-11
ENH: Configuration enhancement (search, value sort fields, allow raw data storage)

mfendeksilverstripe:
# ENH: Configuration enhancement (search, value sort fields, allow raw data storage)

* added the ability to specify search, value and sort fields which were previously hard coded
* added the ability to store raw data to cover the cases where data serialisation is done on the model level as opposed to the controller level (form)

This change is mostly code reuse from one of my projects so I would be keen on getting early feedback before sinking too much effort into this (docs, unit tests, backwards compatibility...). I'm mostly interested to first validate if we need to make this change. Here are some topics I would like to cover:

* responsibility of this form field (should serialisation be covered or not)
* "correct setup" detection (some of the code is making assumptions instead of forcing the developer to properly configure the field)

## Backwards compatibility

* some field setups may break without an update

### DIFF
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -49,9 +49,47 @@ class TagField extends MultiSelectField
     protected $canCreate = true;
 
     /**
+     * This is the field that populates the label displayed in the UI
+     * It can be either a DB field or a model method name
+     *
      * @var string
      */
     protected $titleField = 'Title';
+
+    /**
+     * This is the field that is used to store selected values
+     * It has to be a DB field or null
+     * Use null for the auto-detection
+     *
+     * @var string
+     */
+    protected $valueField = 'Title';
+
+    /**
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Sorry this hasn't been reviewed until now. The concept seems sensible. Are you still interested in having this change merged in?
If so, please retarget to (and rebase on) the `3` branch and add appropriate test coverage for the new API.

GuySartorelli:
It seems like there's going to be no further activity on this pull request, so we’ve closed it for now. Please open a new pull-request if you want to re-approach this work, or for anything else you think could be fixed or improved.

mfendeksilverstripe:
@GuySartorelli Replaced by https://github.com/silverstripe/silverstripe-tagfield/pull/274

PR #229, opened 2023-01-18
DEP Upgrade frontend build stack

GuySartorelli:
See https://github.com/silverstripe/silverstripe-admin/pull/1389 for a wider discussion on most of the changes

## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/644


GuySartorelli:
Keeping this as draft for now - once https://github.com/silverstripe/webpack-config/pull/59 is released this will need `yarn upgrade @silverstripe/webpack-config` to be run, and its new `yarn.lock` to be committed.

src/TagField.php
```diff
@@ -497,7 +497,7 @@ protected function getOrCreateTag($value)
         }
 
         // Create new instance if not yet saved
-        if ($this->getCanCreate()) {
+        if ($this->getCanCreate() && $value) {
```
ChloeSartorelli, line 500: Without this, a new empty value entry was created.

GuySartorelli:
Webpack config changes have been tagged and included in this PR

emteknetnz (changes requested):
CSS is wrong, it looks different from CMS 4

This PR:
image

CMS 4:
image

Other issues:
- When creating a new tag, pressing 'enter' submits the form, when it should instead create the new tag
- I cannot create new tags after saving a page with tags.

GuySartorelli:
For styling please compare this with the `TreeMultiselectField`, not with the CMS 4 version. We went through multiple major versions of react-select and it changed a lot of the markdown and styles that came out of the box. I think some leeway for minor changes in styling is acceptable especially given that this is a major upgrade - people will expect things to look a little different. The styling is _different_ but it's not breaking anything by being different.

GuySartorelli:
> When creating a new tag, pressing 'enter' submits the form, when it should instead create the new tag

I can't reproduce this. Can you please include steps for reproduction (or just show me it happening since I sit next to you)

> I cannot create new tags after saving a page with tags.

Interesting.... I had tested this but maybe I changed something after testing it. For me, it's not that I can't create new tags specifically but I can't select multiple tags. I will resolve this.

client/src/components/TagField.js
```diff
@@ -124,7 +175,7 @@ class TagField extends Component {
     }
 
     // if this is a single select then we just need the first value
-    if (!passThroughAttributes.multi && passThroughAttributes.value) {
+    if (!multi && passThroughAttributes.value) {
```
ChloeSartorelli, line 178: This isn't a passthrough anymore - missed this the first attempt when I pulled multi out as a distinct variable.

emteknetnz:
> Can you please include steps for reproduction

Not much beyond "add a tag and press enter" I used firefox if that makes a difference

I'll retest once I look at all the resolved PRs again, we'll assume it's a non-issue for now

emteknetnz (approved):

PR #233, opened 2023-02-22
add performReadonlyTransformation

w-huber:
An archived page with defined tags using StringTagField cannot be displayed.

```php
class Page extends SiteTree
{
  private static $db = [
    'Tags' => 'Text'
  ];

  public function getCMSFields()
  {
    $tagField = StringTagField::create('Tags', 'Tags')->setCanCreate(true);
    $fields->addFieldToTab('Root.Main', $tagField, 'Content');
    return $fields;
  }
}
```
```
[Emergency] Uncaught TypeError: array_key_exists(): Argument #1 ($key) must be a valid array offset type
GET /admin/pages/edit/show/3

Line 32 in /var/www/html/vendor/silverstripe/framework/src/Forms/SingleLookupField.php
```

```
23     /**
24      * @return mixed|null
25      */
26     protected function valueToLabel()
27     {
28         $value = $this->value;
29         $source = $this->getSource();
30         $source = ($source instanceof Map) ? $source->toArray() : $source;
31 
32         if (array_key_exists($value, $source ?? [])) {
33             return $source[$value];
34         }
35 
36         return null;
37     }
38
```

Adding this code to silverstripe/tagfield/src/StringTagField.php solves the problem...

```php
class StringTagField extends DropdownField
{
...
  {
    /** @var ReadonlyField $copy */
    $copy = $this->castedCopy(ReadonlyField::class);
    $copy->setValue(implode(', ', $this->Value()));

    return $copy;
  }
...
}
```

### DIFF
--- a/src/StringTagField.php
+++ b/src/StringTagField.php
@@ -412,4 +412,11 @@ class StringTagField extends DropdownField
 
         return $this;
     }
+
+    public function performReadonlyTransformation()
+    {
+        $field = parent::performReadonlyTransformation();
+        $field->setValue(implode(', ', $this->Value()));
+        return $field;
+    }
 }
--- a/tests/StringTagFieldTest.php
+++ b/tests/StringTagFieldTest.php
@@ -171,22 +171,26 @@ class StringTagFieldTest extends SapphireTest
         $this->assertNotEmpty($attributes['data-schema']);
     }
 
-    /**
-     * Ensure a source of tags matches the given string tag names
-     *
-     * @param array $expected
-     * @param DataList $actualSource
-     */
-    protected function compareTagLists(array $expected, DataList $actualSource)
+    public function testPerformReadonlyTransformation()
     {
-        $actual = array_keys($actualSource->map('ID', 'Title')->toArray());
-        sort($expected);
-        sort($actual);
+        $field = new StringTagField('Tags');
+        $field->setSource(['Test1', 'Test2', 'Test3']);
 
-        $this->assertEquals(
-            $expected,
-            $actual
-        );
+        // Ensure a single value can be rendered
+        $field->setValue(['Test2']);
+        $field_readonly = $field->performReadonlyTransformation();
+        $this->assertEquals('Test2', $field_readonly->Value());
+
+        // Ensure multiple valid values are rendered
+        $field->setValue(['Test1', 'Test2']);
+        $field_readonly = $field->performReadonlyTransformation();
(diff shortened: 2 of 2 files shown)


michalkleiner:
Thanks @w-huber, that's a good start for the PR. Can you check the test failure and fix the code formatting so the test passes?

w-huber:
I saw that there was already a similar pull request in February 2020. ( #156 )
This also has an implemented test.

michalkleiner:
Ah, good spot. We should probably get that one over the line since it has a test.

GuySartorelli:
Given the comments above, sounds like this is a duplicate.
Closing in favour of #156

GuySartorelli:
Actually on further inspection, this is a _much_ cleaner solution. I'm going to reopen this, make a small tweak, and add some tests, and then I'll close the other PR.

tests/StringTagFieldTest.php
```diff
@@ -171,22 +171,26 @@ public function testSchemaIsAddedToAttributes()
         $this->assertNotEmpty($attributes['data-schema']);
     }
 
-    /**
-     * Ensure a source of tags matches the given string tag names
-     *
-     * @param array $expected
-     * @param DataList $actualSource
-     */
-    protected function compareTagLists(array $expected, DataList $actualSource)
```
ChloeSartorelli, line 180: The `compareTagLists` method isn't used at all - it's a carry-over from `TagFieldTest.php`.
Deleted because even if it _was_ used it would fail, since there's no `use SilverStripe\ORM\DataList` statement in the file.

src/StringTagField.php
```diff
@@ -412,4 +412,11 @@ public function setCanCreate($canCreate)
 
         return $this;
     }
+
+    public function performReadonlyTransformation()
+    {
+        $field = parent::performReadonlyTransformation();
+        $field->setValue(implode(', ', $this->Value()));
+        return $field;
+    }
```
ChloeSartorelli, line 421: This PR originally used a `ReadOnlyField` - but this respects the parent classes and only makes the change that is necessary - setting the value as a string.

ChloeSartorelli: Note that this is preferred over https://github.com/silverstripe/silverstripe-tagfield/pull/156 because it's _way_ simpler, and it doesn't add an entirely new class (which means both a smaller maintenance surface and we can ship this fix in a patch release)

michalkleiner (approved):

GuySartorelli:
Approved by a core committer and was told by another to just self merge, so.... self merging

PR #240, opened 2023-04-21
MNT Use React Testing Library

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1419


client/src/components/TagField.js
```diff
@@ -231,6 +242,10 @@ TagField.defaultProps = {
   lazyLoad: false,
   creatable: false,
   multi: false,
+  SelectComponent: Select,
```
emteknetnz, line 249: Allowing passing in components to because we're no longer doing shallow() renders and the react-select components don't want to simply render, instead they throw all sorts of errors on render() including `Warning: Invalid hook call. Hooks can only be called inside of the body of a function component.`

client/src/components/tests/TagField-test.js
```diff
@@ -1,88 +1,64 @@
...
-      props.lazyLoad = true;
-      const wrapper = shallow(
-        <TagField {...props} />
-      );
-      expect(wrapper.find(AsyncCreatableSelect).length).toBe(1);
-    });
-  });
+import { render, screen } from '@testing-library/react';
 
-  describe('with lazyLoad on and given a URL', () => {
```
emteknetnz, line 63: It no longer makes sense to test a the fetch request here using a mock as the AsyncSelectComponent we're passing in doesn't make a fetch

client/src/components/TagField.js
ChloeSartorelli: Please also add prop types for these.

package.json
```diff
@@ -28,6 +28,7 @@
     "lint-sass": "sass-lint client/src"
   },
   "dependencies": {
+    "@testing-library/react": "^14.0.0",
```
ChloeSartorelli: This should be a dev dependency

GuySartorelli (changes requested):

client/src/components/TagField.js
emteknetnz: Updated

package.json
emteknetnz: Updated

GuySartorelli (approved):

PR #244, opened 2023-05-24
MNT Support Storybook v7

sabina-talipova:
### Description
Updates for `@storybook/addon-knobs` to support Storybook v7:
- Replaced `storiesOf` API with CSF(Component Story Format). [See notice](https://storybook.js.org/docs/react/migration-guide#storiesof-support-discontinued-by-default);
### Parent issue
- https://github.com/silverstripe/silverstripe-cms/issues/2842


package.json
```diff
@@ -34,6 +34,7 @@
     "prop-types": "^15.8.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-select": "^5.7.3",
```
emteknetnz, line 37: Why are we adding react-select here? We haven't changed any frontend code?

emteknetnz (changes requested):
Please change the UPD prefix on the commit and pull-request to MNT instead - https://docs.silverstripe.org/en/5/contributing/code/#commit-messages

package.json
sabina-talipova: I assume that we accidentally have removed this dependency as part of [this PR](https://github.com/silverstripe/silverstripe-tagfield/pull/229), but we use `Select` component [here](https://github.com/silverstripe/silverstripe-tagfield/blob/3/client/src/components/TagField.js#L2).

emteknetnz (approved):
We'll create a separate card to deal with the 4 space indent which should be a 2 space indent

Issue #247, opened 2023-06-28
TagField triggers edit form change state when there are no changes


obj63mc:
Looks like with SS5 release, related to the old issue #125. Currently on a site using a tag field, whenever a cms page is loaded - it thinks there are changes that need to be saved.

Example on a pages getCMSFields function - 

```php
$this->beforeUpdateCMSFields(function ($fields) {
    $fields->addFieldToTab('Root.Main', TagField::create(
        'ProductPages',
        'Products',
        ProductPage::get(),
        $this->ProductPages()
    )->setCanCreate(false)->setShouldLazyLoad(true), 'Content');
});
```

When I load any page in the CMS that currently has a value set - the Save/Publish buttons are lit up as green post page load causing the form to think the edit state is changed, thus when clicking off it asks if you want to leave due to unsaved changes.  If I comment out this field all works as normal.

## PR
- https://github.com/silverstripe/silverstripe-tagfield/pull/251

GuySartorelli:
PR merged and tagged as [3.0.1](https://github.com/silverstripe/silverstripe-tagfield/releases/tag/3.0.1)

PR #251, opened 2023-07-17
FIX TagField triggers edit form change without changes

sabina-talipova:
## Description
The tag field input works differently than a simple text field and does not retain the new value after selecting or creating a new tag. Therefore, it is necessary to track the tag itself rather than typing in the input field. An additional method has been added that is called when a new tag is added or created.

Steps to reproduce the problem:
- Create a new Blog and add a new Blog post to it (easy way to get a tag field).
- Go to Post options tab.
- Start typing in the Categories or Tags field.
- The Save button is highlighted in color, which indicates that there are changes on the page.
- When you try to go to another page, a message about unsaved data is displayed.
- Even if the text is removed from the field, the message about unsaved data and the highlighted Save button remain.

Steps to test:
- Create a new Blog and add a new Blog post to it.
- Go to Post options tab.
- Start typing in the Categories or Tags field.
- The Save button didn't change.
- Select suggested option in dropdown list.
- New tag was added.
- If there were not any tags initially, if you add and remove tag, Save button isn't  highlighted. 
- The Save button is highlighted in color, which indicates that there are changes on the page.

## Parent issue
- https://github.com/silverstripe/silverstripe-tagfield/issues/247


client/src/legacy/entwine/TagField.js
```diff
@@ -54,7 +62,7 @@ window.jQuery.entwine('ss', ($) => {
         // clone the object (so we don't modify the original),
         opts = $.extend({}, opts);
         // modify it,
-        opts.ignoreFieldSelector += ', .ss-tag-field .Select :input';
```
sabina-talipova, line 57: `.Select` class doesn't exist in DOM

package.json
```diff
@@ -34,6 +34,7 @@
     "prop-types": "^15.8.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-select": "^5.7.3",
```
ChloeSartorelli:
```suggestion
    "react-select": "^5.5.8",
```
We must match the constraint in silverstripe/admin if we're going to add this at all.

client/src/components/TagField.js
```diff
@@ -55,6 +56,16 @@ class TagField extends Component {
    * @param {string} value
    */
   handleChange(value) {
+    let hasChanges = false;
+
+    if (JSON.stringify(this.state.initalState) !== JSON.stringify(value)) {
+          hasChanges = true;
+    }
```
ChloeSartorelli:
```suggestion
    if (JSON.stringify(this.state.initalState) !== JSON.stringify(value)) {
      hasChanges = true;
    }
```

GuySartorelli (changes requested):
This works in an entwine context (such as on a blog page) - but it _doesn't_ work in a react context (such as when the `TagField` is inside an elemental block).

There are two ways you could approach this from here:
1. Just reinstate the CMS 4 behaviour - i.e. tagfield input fields are _always_ ignored.
    - Remove all of the changes you've made
    - in the `client/src/legacy/entwine/TagField.js` file make this change:
      ```diff
      - opts.ignoreFieldSelector += ', .ss-tag-field .Select :input';
      + opts.ignoreFieldSelector += ', .ss-tag-field :input';
      ```
1. Make this correctly detect changes in react
    - Basically do what you have done, but don't add "no-change-track" in PHP or in the entwine code
    - Handle all of the "no-change-track" classname stuff inside the react component itself
    - Don't add "no-change-track" to the _holder_ element at all, and in fact remove the `getChangeTrackerOptions()` section. Instead, just add or remove the "no-change-track" classname from the appropriate input component directly
    - Note that there are _two_ `:input` fields in the react select component. You want to add the classname to both of them initially, and only remove it from the one that holds the actual value, not the one you type into.
    - you may need to refer to [the react select docs](https://react-select.com/styles#cx-and-custom-components) for how to add class names to specific components within the select component.
    
Obviously option 1 is way easier, but option two is a slightly better user experience. I'll leave it to you to choose which you want to try.

package.json
sabina-talipova: We have 2 different versions now:
`"silverstripe-cms"`: `"react-select": "^5.6.0"`,
`"silverstripe-admin"`: `"react-select": "^5.5.8"`

Probably, I'll update `cms` and `admin` to `"react-select": "^5.7.3"` to be consistent.
I assume that is the best practice to explicitly require all dependencies that we use in module.
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md

sabina-talipova:
> Just reinstate the CMS 4 behaviour - i.e. tagfield input fields are always ignored.

I think it's bad idea. I suppose that it was made by mistake. Our user expect that all fields track changes, but it's annoying when they track changes for some field with ability to select options when user just start to typing.

client/src/components/TagField.js
sabina-talipova: Updated.

sabina-talipova:
> Note that there are two :input fields in the react select component. You want to add the classname to both of them initially, and only remove it from the one that holds the actual value, not the one you type into.

`changetracker` doesn't care about any input fields if they are not active.

package.json
sabina-talipova: Updated

sabina-talipova:
>  but it doesn't work in a react context (such as when the TagField is inside an elemental block).

There are more deep problem then it can be covered by this task. Using TagFields in Elemental Area doesn't work at all. 
I created new ticket for this. 
- https://github.com/silverstripe/silverstripe-tagfield/issues/252

GuySartorelli (approved):
Works locally.

At some point in the future we'll probably want a cleaner approach we can use across multiple react components but for now since this is the only one that appears to need such a workaround, this will do just fine.

Issue #252, opened 2023-07-19
TagField used in Elemental Area loses its value on blur


sabina-talipova:
## Description
TagField used in Elemental Area does not save created or added values. This issue has beed tested in v5 and v5.0.

## Steps to reproduce
- Create new Block with TagFileld;
For example, `$fields->addFieldToTab('Root.Main', TagField::create('Tags', 'Tags', BlogTag::get()));`
- Add Block to a page;
- Add new tag to the tag field;
- Click out of tag field;
- New tag disappeared;

### Related issue
- https://github.com/silverstripe/silverstripe-tagfield/issues/247

GuySartorelli:
Just as a sanity check, did you also add an appropriate relation on the block for it to save into?

GuySartorelli:
I've also tested this in 4.13 which also has this problem.

GuySartorelli:
I've tried this again in CMS 4.13 and wasn't able to reproduce so looks like it's fine there.
Merged the PR for 5.0. It'll be automatically tagged by github actions.

PR #256, opened 2023-08-18
Add onBlur handler back

### LINKED ISSUE #57: CMS5 Inline editing loses value on blur
So I think the underlying issue is from https://github.com/silverstripe/silverstripe-tagfield/issues/252 but just so there is a record here too - the tagfield loses its value on blur so it is not possible to use it from the the elemental area inline editor

blueo:
was removed in https://github.com/silverstripe/silverstripe-tagfield/commit/249842fab884d33e4fc3177ebe42c65bb648c164

i've confirmed it fixes https://github.com/dnadesign/silverstripe-elemental-virtual/issues/57  https://github.com/silverstripe/silverstripe-tagfield/issues/252

and I checked it doesn't appear to affect the change tracker state: https://github.com/silverstripe/silverstripe-tagfield/pull/251


GuySartorelli (approved):
Nice and simple, works well locally. Thanks!

Issue #270, opened 2023-12-13
SS5 z-index of dropdown behind TinyMCE


christopherdarling:
**SS5 Screenshot**

![image](https://github.com/silverstripe/silverstripe-tagfield/assets/178039/b45bb8dc-380a-4df6-873a-dfb1f6429e9d)

**SS4 Screenshot**

![image](https://github.com/silverstripe/silverstripe-tagfield/assets/178039/5c44cf60-69ab-46e6-bd6d-d7ae3310f858)

Increasing the z-index of the dropdown seems to fix the issue as `.tox .tox-editor-header` has a z-index of 1 also.

## PR
- https://github.com/silverstripe/silverstripe-tagfield/pull/271
- https://github.com/silverstripe/silverstripe-tagfield/pull/272

michalkleiner:
Thanks for the report @christopherdarling, do you wanna try to create a PR with a fix?

christopherdarling:
@michalkleiner looks like https://github.com/JedWatson/react-select/blob/06e34882638d1526b9f5a1238bb567a3e9460ce5/packages/react-select/src/components/Menu.tsx#L292 needs changing, I have had a look through the docs but can't see a way of updating this. 

There is an open ticket (https://github.com/JedWatson/react-select/issues/5329) and a few closed ones on this subject but doesn't seem to be a solution so might be a style we need to override through the tagfield styles. I'll send a PR later/soon.

christopherdarling:
lint wasn't working on my local set up so you may want to re-build this but this seems to work; https://github.com/silverstripe/silverstripe-tagfield/pull/271

christopherdarling:
new PR @ https://github.com/silverstripe/silverstripe-tagfield/pull/272 @GuySartorelli (with a better description too 😉 )

GuySartorelli:
Thanks for reporting and fixing that! The patch will be automatically tagged by GitHub actions

PR #271, opened 2023-12-13
FIX z-index of select dropdown clashing with TinyMCE toolbar

christopherdarling:
https://github.com/silverstripe/silverstripe-tagfield/issues/270


GuySartorelli:
Please provide some explanation of what the specific problem is that this pr intends to resolve.

christopherdarling:
@GuySartorelli please see the linked issue for context

GuySartorelli:
If there is an issue that this pr relates to, please at the very very least link to it in the pr description as per [the documented contribution guide](https://docs.silverstripe.org/en/5/contributing/code/#create-the-pr)

christopherdarling:
Sure all done

client/src/components/TagField.scss
```diff
@@ -108,3 +108,7 @@
     color: #0071e6;
   }
 }
+
+.ss-tag-field__menu {
+  z-index: 10;
```
ChloeSartorelli, line 115:
```suggestion
  // Prevents the dropdown from hiding behind TinyMCE
  // see https://github.com/JedWatson/react-select/issues/5329
  z-index: 10;
```
Please add this so there's some immediate feedback about why this is needed - it's pretty weird to need a z-index like this so we're bound to want to know about it in the future.

GuySartorelli (changes requested):
Looks good, works locally. Just one small change.

If you want this to be available immediately as a patch release, please retarget the PR to the `3.1` branch (you may need to [reset your commits](https://gist.github.com/GuySartorelli/74627843aeb75e12b777b1f44cb22d2f) afterward).

PR #274, opened 2023-12-21
ENH: Configuration enhancement (search, value sort fields, allow raw data storage)

mfendeksilverstripe:
# ENH: Configuration enhancement (search, value sort fields, allow raw data storage)

* added the ability to specify search, value and sort fields which were previously hard coded
* added the ability to store raw data to cover the cases where data serialisation is done on the model level as opposed to the controller level (form)

This change is mostly code reuse from one of my projects so I would be keen on getting early feedback before sinking too much effort into this (docs, unit tests, backwards compatibility...). I'm mostly interested to first validate if we need to make this change. Here are some topics I would like to cover:

* responsibility of this form field (should serialisation be covered or not)
* "correct setup" detection (some of the code is making assumptions instead of forcing the developer to properly configure the field)

### Backwards compatibility considerations

* changes should be backwards compatible

## Related issues

- https://github.com/silverstripe/silverstripe-tagfield/issues/275

### DIFF
--- a/src/TagField.php
+++ b/src/TagField.php
@@ -49,9 +49,47 @@ class TagField extends MultiSelectField
     protected $canCreate = true;
 
     /**
+     * This is the field that populates the label displayed in the UI
+     * It can be either a DB field or a model method name
+     *
      * @var string
      */
     protected $titleField = 'Title';
+
+    /**
+     * This is the field that is used to store selected values
+     * It has to be a DB field or null
+     * Use null for the auto-detection
+     *
+     * @var string
+     */
+    protected $valueField = 'Title';
+
+    /**
+     * This is the field which drives the "suggest" action via text-based search
+     * It has to be a DB field
+     *
+     * @var string
+     */
+    protected $searchField = 'Title';
+
+    /**
+     * This is the field which drives the order of results that appear in the "suggest" action via text-based search
+     * It has to be a DB field or empty string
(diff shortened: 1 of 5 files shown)


GuySartorelli:
Please create an issue, rather than a pull request, for discussions about this sort of thing (as per our [contribution guide](https://docs.silverstripe.org/en/5/contributing/code/))

Please include as much information as possible about the intended use case in the issue that you create - as it is, I'm not sure what you mean by "data serialisation is done on the model level as opposed to the controller level (form)" for example.

I'm also marking this as draft since it's not intended to be merged as-is.

mfendeksilverstripe:
@GuySartorelli Replaces https://github.com/silverstripe/silverstripe-tagfield/pull/227

mfendeksilverstripe:
@GuySartorelli I've implemented your feedback from the previous version of this PR, please review.

GuySartorelli:
@mfendeksilverstripe Based on the description it sounds like you're more after a discussion than a peer review on this PR, is that correct? If so, please see https://github.com/silverstripe/silverstripe-tagfield/pull/274#issuecomment-1866911081 for how to better get that discussion going.

I don't have time to review this PR this side of Christmas, so if it is just a review of this PR as it is, please update the description to indicate that and for best results remind me when business starts again next year.

mfendeksilverstripe:
Hey @GuySartorelli I think I've addressed all your feedback. Can you please have another look?

Issue #309, opened 2024-09-27
ReadonlyTagField loads all option items, without lazy loading, prior to returning selected ones


adunn49:
### Module version(s) affected

3.2.1

### Description

Hey team,
I believe there is a performance issue when loading the TagField in read only mode (using TagField 3.2.1). The finding of this issue has come about through the combination of a number of other modules, namely [a virtual elemental block](https://github.com/dnadesign/silverstripe-elemental-virtual) (that now uses TagField), in conjunction with [a workflow module](https://github.com/symbiote/silverstripe-advancedworkflow) that renders the TagField in read only mode. In a project this resulted in tens of thousands of blocks being returned in a query, causing very slow behaviour in the CMS.

### How to reproduce

This issue can be seen with the following steps:
* Adding a TagField to a page
* Making the source of the TagField pages within the site
* Adding a read only transformation to the added field.

So, in code, something like …

````
private static array $has_one = [
    'Tags' => TagField::class,
];

/**
 * @inheritDoc
 */
public function getCMSFields()
{
    $fields = parent::getCMSFields();

    $availablePages = SiteTree::get();

    $fields->addFieldToTab(
        'Root.Main',
        TagField::create('TagsID', 'Linked Page', $availablePages)
            ->setIsMultiple(false)
            ->setCanCreate(false)
            ->setShouldLazyLoad(true)
            ->setTitleField('Title')
            ->performReadonlyTransformation()
    );

    return $fields;
}
````

With these steps, and using debugging tools, although you won't experience the levels of performance degredation without a large number of items, you should be able to see that this loads all pages and performs a function on each to return an array of options (https://github.com/silverstripe/silverstripe-tagfield/blob/3/src/TagField.php#L310-L328).

### Possible Solution

The issue, I believe, stems from the ReadOnlyTagField class (https://github.com/silverstripe/silverstripe-tagfield/blob/3/src/ReadonlyTagField.php#L30) where the `getOptions` method is called on the TagField class. Looking in the getOptions function it can be seen that without the `$onlySelected` attribute passed in, lazy loading is assumed false and this will load all items. It would seem that ReadOnlyTagField class only needs the selected fields so calling `getOptions(true) `might be a solution, but not sure it is as simple as that!?!

### Validations

- [X] Check that there isn't already an issue that reports the same bug
[...]

PR #310, opened 2024-10-01
ENH Refactor the way selected options are returned for read only tag fields

adunn49:
## Description
When transforming a tag field to read only mode and calling the `Field` method on it for rendering, it was observed that all options in the tag field source were being returned from the database, and iterated to provide a full list on which the selected options are then filtered. Where there are large numbers of items in the source this was causing performance issues.

The `getOptions` method that is called for setting the values on the read only tag field accepts a `$onlySelected` parameter  which, when set to `true`, will avoid the loading of all options, and instead just return the selected values. Calling the getOptions method in this way means that we no longer need the filtering within the `ReadonlyTagField` class.

## Manual testing steps
These steps are included on issue #309 

## Issues
- #309 

## Pull request checklist
- [x] The target branch is correct
    - See [picking the right version](https://docs.silverstripe.org/en/contributing/code/#picking-the-right-version)
- [x] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
    - Small amounts of additional linting are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [x] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [x] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] This change is covered with tests (or tests aren't necessary for this change)
- [x] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [ ] CI is green

### DIFF
--- a/src/ReadonlyTagField.php
+++ b/src/ReadonlyTagField.php
@@ -27,7 +27,8 @@ class ReadonlyTagField extends TagField
     {
         $options = array();
 
-        foreach ($this->getOptions()->filter('Selected', true) as $option) {
+        // only get the selected options
+        foreach ($this->getOptions(true) as $option) {
             $options[] = $option->Title;
         }
 
--- a/tests/TagFieldTest.php
+++ b/tests/TagFieldTest.php
@@ -408,6 +408,22 @@ class TagFieldTest extends SapphireTest
         $field->setTitleField('Name');
         $readOnlyField = $field->performReadonlyTransformation();
         $this->assertEquals('Name', $readOnlyField->getTitleField());
+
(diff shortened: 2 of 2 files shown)


GuySartorelli (approved):
Very sensible change and good test coverage, thank you.

PR #311, opened 2024-10-31
API Disable ancestor FieldValidators

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11422

### DIFF
--- a/src/StringTagField.php
+++ b/src/StringTagField.php
@@ -7,13 +7,13 @@ use SilverStripe\Control\Controller;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Forms\DropdownField;
-use SilverStripe\Forms\Validator;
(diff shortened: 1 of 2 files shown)


src/StringTagField.php
```diff
@@ -26,6 +26,12 @@
  */
 class StringTagField extends DropdownField
 {
+    private static array $field_validators = [
+        // Disable validation as StringTagField is a special case where labels can be the values
+        // rather than the keys always being the values - for instance when creating new tags
+        OptionFieldValidator::class => null,
```
GuySartorelli, line 32: This means this field now has _no_ validation, where it used to have validation through inheritance. That's no good.

Same with TagField.

GuySartorelli (changes requested):

GuySartorelli (approved):
LGTM, sucks these don't have validation but that's what the status quo was before so thus it shall remain

GuySartorelli (approved):
LGTM

PR #312, opened 2024-11-15
FIX Avoid infinite recursive loop with attributes in schemadata

GuySartorelli:
Reflects changes in https://github.com/silverstripe/silverstripe-framework/pull/11469 - that may need to be merged for CI to go green.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/9124

### DIFF
--- a/tests/TagFieldTest.php
+++ b/tests/TagFieldTest.php
@@ -413,6 +413,8 @@ class TagFieldTest extends SapphireTest
         $field = new TagField('Tags', '', TagFieldTestBlogTag::get());
         $field->setTitleField('Title');
         $field->setValue(['Tag1']);
+        $form = new Form();
+        $field->setForm($form);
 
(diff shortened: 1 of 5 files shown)


tests/TagFieldTest.php
```diff
@@ -413,6 +413,8 @@ public function testReadonlyTransformation()
         $field = new TagField('Tags', '', TagFieldTestBlogTag::get());
         $field->setTitleField('Title');
         $field->setValue(['Tag1']);
+        $form = new Form();
+        $field->setForm($form);
```
ChloeSartorelli, line 417: Without this we get the following error due to calling `getSchemaState()` which we didn't used to call:

> LogicException: Field must be associated with a form to call Link(). Please use $field->setForm($form);

emteknetnz (approved):

PR #317, opened 2025-01-14
API Rename FormField Value to getFormattedValue

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11468

### DIFF
--- a/src/StringTagField.php
+++ b/src/StringTagField.php
@@ -173,7 +173,7 @@ class StringTagField extends DropdownField
                 'lazyLoad' => $this->getShouldLazyLoad(),
                 'creatable' => $this->getCanCreate(),
                 'multi' => $this->getIsMultiple(),
-                'value' => $this->formatOptions($this->Value()),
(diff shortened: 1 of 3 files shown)


src/StringTagField.php
```diff
@@ -384,7 +384,7 @@ public function setCanCreate($canCreate)
     public function performReadonlyTransformation()
     {
         $field = parent::performReadonlyTransformation();
-        $field->setValue(implode(', ', $this->Value()));
+        $field->setValue(implode(', ', $this->getFormattedValue()));
```
ChloeSartorelli: Shouldn't we be setting the internal value of the readonly field to the internal value of this field?

GuySartorelli (changes requested):

src/StringTagField.php
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #336, opened 2025-11-05
Convert TagField to functional component


emteknetnz:
TagField.js is a react class component that needs to be converted to a functional component using hooks

## Acceptance Criteria
- [ ] TagField.js is converted to a functional component using hooks.
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved.

## PRs
- https://github.com/silverstripe/silverstripe-tagfield/pull/337

GuySartorelli:
PR merged

PR #337, opened 2025-11-06
ENH Refactor TagField from class to functional component

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-tagfield/issues/336

Can test with the following:

```php
// app/src/BlogPostAdmin.php

use SilverStripe\Admin\ModelAdmin;

class BlogPostAdmin extends ModelAdmin
{
    private static string $url_segment = 'blog-post-admin';
    private static string $menu_title = 'Blog Post Admin';
    private static array $managed_models = [
        BlogPost::class,
        BlogTag::class,
    ];
}

// app/src/BlogPost.php

use SilverStripe\Forms\FieldList;
use SilverStripe\ORM\DataObject;
use SilverStripe\TagField\TagField;

class BlogPost extends DataObject
{
    private static string $table_name = 'BlogPost';
    private static string $singular_name = 'Blog Post';
    private static string $plural_name = 'Blog Posts';

    private static array $db = [
        'Title' => 'Varchar(255)',
        'Content' => 'Text',
    ];

    private static array $many_many = [
        'BlogTags' => BlogTag::class
    ];

    public function getCMSFields(): FieldList
    {
        $field = TagField::create(
            'BlogTags',
            'Blog Tags',
            BlogTag::get(),
            $this->BlogTags()
        )
            ->setShouldLazyLoad(true)
            ->setCanCreate(true);
        $fields->addFieldToTab('Root.Main', $field);
        return $fields;
    }
}

// app/src/BlogTag.php

use SilverStripe\ORM\DataObject;

class BlogTag extends DataObject
{
    private static string $table_name = 'BlogTag';

    private static array $db = [
        'Title' => 'Varchar(200)',
    ];

    private static array $belongs_many_many = [
        'BlogPosts' => BlogPost::class
    ];

    private static string $singular_name = 'Blog Tag';
    private static string $plural_name = 'Blog Tags';

    public function requireDefaultRecords(): void
    {
        parent::requireDefaultRecords();
        $tags = [
            'Lorem Ipsum',
            'Dolor Sit Amet',
            'Consectetur Adipiscing',
            'Elit Sed Do',
            'Eiusmod Tempor',
            'Incididunt Ut Labore',
            'Et Dolore Magna Aliqua',
            'Ut Enim Ad Minim',
        ];
        foreach ($tags as $title) {
            if (!self::get()->filter('Title', $title)->exists()) {
                $tag = self::create();
                $tag->Title = $title;
                $tag->write();
            }
        }
    }
}
```


client/src/components/TagField.js
```diff
@@ -51,205 +73,155 @@ class TagField extends Component {
...
+  }, [lazyLoad, options, fetchOptionsDebounced]);
+
+  /**
+   * Determine if this input should be "controlled" or not. Controlled inputs should rely on their
+   * value coming from props and a change handler provided to update the state stored elsewhere.
+   * This is specifically the case for use with `redux-form`.
+   *
+   * @return {boolean}
+   */
+  const isControlled = useCallback(() => typeof onChange === 'function', [onChange]);
```
emteknetnz: Has to move this because it's used in handleChange()

client/src/components/TagField.js
```diff
@@ -11,38 +11,60 @@ import url from 'url';
...
-    }
-  }
+  /**
+   * Debounced function to fetch options from the server, optionally using the given string as a filter.
+   * The actual fetch is delayed by 500ms to avoid excessive requests while the user is typing.
+   *
+   * @param {string} input
+   * @return {Promise<Array<Object>>}
+   */
+  const fetchOptionsDebounced = useCallback(
```
emteknetnz: Had to move up as it's used by getOptions()

client/src/components/TagField.js
```diff
@@ -51,205 +73,155 @@ class TagField extends Component {
...
-      .then((response) => response.json())
-      .then((json) => json.items.map(
-        (item) => ({
-          [labelKey]: item.Title,
-          [valueKey]: item.Value,
-          Selected: item.Selected,
-        })
-      ));
-  }
+  const valueInOptions = (checkValue, checkOptions, key) => {
```
emteknetnz, line 125: Had to move up as is used by isValidNewOption()

client/src/components/TagField.js
```diff
@@ -51,205 +73,155 @@ class TagField extends Component {
...
-    if (!multi && passThroughAttributes.value) {
-      if (Object.keys(passThroughAttributes.value).length > 0) {
-        const value =
-          passThroughAttributes.value[
-            Object.keys(passThroughAttributes.value)[0]
-          ];
+  // if this is a single select then we just need the first value
+  if (!multi && passThroughAttributes.value) {
+    if (Object.keys(passThroughAttributes.value).length > 0) {
+      const singleValue =
```
emteknetnz, line 187: Renamed from `value` as that's now a prop var

client/src/components/TagField.js
```diff
@@ -51,205 +73,155 @@ class TagField extends Component {
...
+      }
+    }
+    return false;
+  };
 
   /**
    * Check if a new option can be created based on a given input
    * @param {string} inputValue
-   * @param {array|object} value
+   * @param {array|object} checkValue
```
emteknetnz, line 138: Renamed as `value` is prop value

client/src/components/tests/TagField-test.js
```diff
@@ -40,7 +107,7 @@ test('TagField should render a CreatableSelect with creatable option', () => {
   expect(container.querySelector('.test-creatable-select')).not.toBeNull();
 });
 
-test('Tagfiled should render an AsyncSelect with lazy load option', () => {
+test('Tagfield should render an AsyncSelect with lazy load option', () => {
```
emteknetnz, line 110: Fixing old type

client/src/components/TagField.js
```diff
@@ -51,205 +73,155 @@ class TagField extends Component {
...
+   * @return {boolean}
+   */
+  const isControlled = useCallback(() => typeof onChange === 'function', [onChange]);
 
   /**
    * Handle a change, either calling the change handler provided (if controlled) or updating
    * internal state of this component
    *
-   * @param {string} value
+   * @param {string} newValue
```
emteknetnz, line 94: Renaming as `value` is a prop var

client/src/components/tests/TagField-test.js
ChloeSartorelli: typo of typo lol

client/src/components/TagField.js
```diff
@@ -11,38 +11,60 @@ import url from 'url';
...
+  creatable = false,
+  multi = false,
+  disabled = false,
+  labelKey = 'Title',
+  valueKey = 'Value',
+  SelectComponent = Select,
+  AsyncCreatableSelectComponent = AsyncCreatableSelect,
+  AsyncSelectComponent = AsyncSelect,
+  CreatableSelectComponent = CreatableSelect,
+  value: initialValue,
```
ChloeSartorelli: What's this colon syntax doing here, compared with `=` syntax? And where does this `initialValue` come from? This wasn't in default props.

client/src/components/TagField.js
```diff
@@ -11,38 +11,60 @@ import url from 'url';
...
+  onChange,
+  optionUrl,
+  onBlur,
+  ...passThroughAttributesFromProps
```
ChloeSartorelli, line 29: The old `passThroughAttributes` included `onChange`, `optionUrl`, `onBlur`, and `value` which have been explicitly pulled out above.

client/src/components/TagField.js
```diff
@@ -11,38 +11,60 @@ import url from 'url';
...
+}) => {
+  const selectComponentRef = useRef(null);
+
+  const [initalState] = useState(initialValue || []);
+  const [hasChanges, setHasChanges] = useState(false);
+  const [value, setValue] = useState(initialValue);
```
ChloeSartorelli: Again what is `initialValue`?

ChloeSartorelli: For future historians: We can't use hooks conditionally, so we can't have e.g:
```js
if (!isControlled()) {
    const [value, setValue] = useState(initialValue);
}
```

That means it _has_ to be part of the state in both conditions if it is needed as state in one of the two conditions.

@emteknetnz probably this should be renamed to `controlledValue` or similar to make it clear we should only use this state when `isControlled()` is true?

client/src/components/TagField.js
ChloeSartorelli: We don't have a `fetchOptions` other than this one, so why rename it?
```suggestion
  const fetchOptions = useCallback(
```

client/src/components/TagField.js
```diff
@@ -11,38 +11,60 @@ import url from 'url';
...
-  componentDidUpdate(previousProps, previousState) {
-    if (previousState.hasChanges !== this.state.hasChanges) {
-      const element = this.selectComponentRef.current.inputRef;
-      const event = new Event('change', { bubbles: true });
-      element.dispatchEvent(event);
-    }
-  }
+  /**
+   * Debounced function to fetch options from the server, optionally using the given string as a filter.
+   * The actual fetch is delayed by 500ms to avoid excessive requests while the user is typing.
```
ChloeSartorelli:
```suggestion
   * Initiate a request to fetch options, optionally using the given string as a filter.
   * The actual fetch is delayed by 500ms to avoid excessive requests while the user is typing.
```
Going back to original doc comment but keeping the new "fetch is delayed" part

client/src/components/TagField.js
```diff
@@ -11,38 +11,60 @@ import url from 'url';
...
+   *
+   * @param {string} input
+   * @return {Promise<Array<Object>>}
```
ChloeSartorelli:
```suggestion
   * @return {Promise<{options: Array<Object>}>}
```
This is what it was originally - I assume that hasn't actually changed?

client/src/components/TagField.js
```diff
@@ -51,205 +73,155 @@ class TagField extends Component {
    * @param {string} input
    * @return {Promise<Array<Object>>|Promise<{options: Array<Object>}>}
    */
-  getOptions(input) {
-    const { lazyLoad, options } = this.props;
-
+  const getOptions = useCallback((input) => {
```
ChloeSartorelli: Why are we using `useCallback` all over the place in this refactor? I don't think it was used in the previous two?

client/src/components/TagField.js
```diff
@@ -51,205 +73,155 @@ class TagField extends Component {
...
-      .then((response) => response.json())
-      .then((json) => json.items.map(
-        (item) => ({
-          [labelKey]: item.Title,
-          [valueKey]: item.Value,
-          Selected: item.Selected,
-        })
-      ));
-  }
+  const valueInOptions = (checkValue, checkOptions, key) => {
```
ChloeSartorelli, line 125: Why was `valueKey` renamed to`key`?
Update: Ahh, `valueKey` is the name of a prop.

client/src/components/TagField.js
```diff
@@ -51,205 +73,155 @@ class TagField extends Component {
...
-    // eslint-disable-next-line no-restricted-syntax
-    for (const item of options) {
-      if (value === item[valueKey]) {
-        return true;
-      }
-    }
-    return false;
+  // Update the value to passthrough with the kept state provided this component is not
+  // "controlled"
+  const passThroughAttributes = { ...passThroughAttributesFromProps };
```
ChloeSartorelli: Why are we setting a new variable instead of just naming it `passThroughAttributes` to start with?

GuySartorelli (changes requested):

client/src/components/TagField.js
emteknetnz: Yeah it's a minor optimization that got put it. Doesn't seem like there's much reason to not keep it since things are functioning correctly and it's slightly more performant.

client/src/components/TagField.js
emteknetnz: I had them split at one point. Updated.

client/src/components/TagField.js
ChloeSartorelli: It's more convoluted and makes reviewing more difficult. If it was just this one PR in isolation it'd probably be fine but given there's over a hundred of these to do, the more like-for-like the better. We can always add optimisations _after_ refactoring.

client/src/components/TagField.js
emteknetnz: tl;dr Fancy syntax so that we can use the nicely name `value` as something we can reassign inside the component would otherwise clash with the prop

It's basically shorthand for

```
function TagField(props) {
  const initialValue = props.value; // immutable
  let value = props.value; // mutable - gets reassigned in the functional component
```

We're using the modern `const TagField({...}) {` syntax though so that we can use Javascript default parameters, so we're using object deconstruction and aliasing to achieve the same thing i.e. not using the `props` variable

client/src/components/TagField.js
emteknetnz: onChange, optionUrl, value - are used in the component

onBlur - have remov

client/src/components/TagField.js
emteknetnz: It's essentially props.value

emteknetnz: I think initialValue is a much better name, it's the initialValue when the component is first created

I could rename it to propValue if you wanted

client/src/components/TagField.js
emteknetnz: Updated

client/src/components/TagField.js
emteknetnz: Updated

client/src/components/TagField.js
emteknetnz: It's creating a new variable so that we're modifying a mutable copy of the props object, thereby preserving the immutability of the passThroughAttributesFromProps rest property (changing prop variables can cause side effects like re-renders)

ChloeSartorelli: As per https://github.com/silverstripe/silverstripe-tagfield/pull/337#discussion_r2500989339 we used to have `onChange`, `optionUrl`, and `value` in these pass through props. Should we be adding them to the `passThroughAttributes` object? Especially since the component we're passing these two is injectable.

(this wasn't meant to be added to the existing thread but GitHub in its infinite wisdom decided I can't make a separate thread for this separate discussion on the same line)

client/src/components/TagField.js
ChloeSartorelli: I'm talking about the state. Though the prop should probably _also_ be renamed since "initial" isn't necessarily true of props - props can change throughout the lifecycle of a component.

Ideally both of those variables should be named in a way that makes it clear when each should be used, since they're both ostensibly "the value of the field"

client/src/components/TagField.js
ChloeSartorelli: As above can you please remove that hook?

Or alternatively should we agree that _all_ of the callbacks in functional components should use this callback and open a card to update the ones that aren't going through this process? (Or is there nuance there where some of them shouldn't use the hook?)

The latter will mean I need to do some reading up on how the hook works and what drawbacks or side effects it might have etc so will slow down review of at least this PR if not other refactor PRs - but if you think the optimisation is worth it I'm open to that.

client/src/components/TagField.js
ChloeSartorelli: Thanks for explaining, that makes sense.

GuySartorelli (commented):
Seems to work as expected locally - just some variable naming to square away and possible changes to usage of callback hook as below.

client/src/components/TagField.js
emteknetnz: .value is already added to passThroughAttributes 
onChange is already passed into the `<DynamicSelect>` component

Have added optionUrl to passThroughAttributes

client/src/components/TagField.js
emteknetnz: Removed

client/src/components/TagField.js
emteknetnz: Renamed to propValue and stateValue to make it clear which is which

client/src/components/TagField.js
```diff
@@ -51,205 +67,156 @@ class TagField extends Component {
...
+  if (!isControlled()) {
+    passThroughAttributes.value = stateValue;
+    passThroughAttributes.optionUrl = optionUrl;
```
ChloeSartorelli: I think `optionUrl` should be added whether it's controlled or not (used to be included implicitly due to not being desconstructed out of the `props` variable)

GuySartorelli (changes requested):
Just the one thing now

client/src/components/TagField.js
emteknetnz: Opps, fixed

GuySartorelli (approved):
LGTM

Issue #335, opened 2025-11-04
Unable to remove a relationship when setIsMultiple is false (no cross icon)


RuthAdele:
### Module version(s) affected

3.4.0 and 4.0.2

### Description

If `->setIsMultiple(false)`, you can select an option, but you cannot remove that selection.

I have tested this on 3.4.0 and 4.0.2, and it's the same on both.

<img width="722" height="68" alt="Image" src="https://github.com/user-attachments/assets/ab9209aa-5caa-4347-adb7-b48c53f589f9" />

**Related issue**
This may be related to the fix from https://github.com/silverstripe/silverstripe-tagfield/issues/195
In that Issue, you can see that there used to be a "x" to remove the selection when `->setIsMultiple(false)`

### How to reproduce

**Steps to recreate**
Create a TagField with setIsMultiple to false.
```
TagField::create(
	'Categories',
	'Categories',
	CourseCategory::get(),
	$this->Categories()
)->setIsMultiple(false)
```

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-tagfield/pull/338

GuySartorelli:
PR merged. Will be automatically tagged by GitHub Actions

PR #338, opened 2025-11-26
FIX Allow non-multi TagField to be cleared

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-tagfield/issues/335

I've done this as a bugfix, though it could be argued that it should be done as a minor release since there's a behavioural change

I've added props with a default value for TagField in case we want to add PHP API later on to to control this i.e. -`>setIsClearable(false)`


GuySartorelli (approved):
LGTM, works as expected for `TagField` and `StringTagField`.

